PR #1, opened 2022-09-14
Ensure consistent ordering for packed objects

angelini:
I discovered this while working on the recent DL slowdown.

We were not properly ensuring that a directory with the same files would always pack in the same order - and result in the same hash key. We intrinsically relied on the order the files were provided to us by the client.

`fsdiff` provides the files to us in an ordered fashion, but when we updated the client to upload files in parallel that ordering was no longer guaranteed on the server side.

This changes ensure the updates are once again ordered and adds a test to verify that this is the case.

### DIFF
--- a/internal/db/tar.go
+++ b/internal/db/tar.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"sort"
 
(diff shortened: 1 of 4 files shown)


scott-rc (approved):
Nice catch!

PR #3, opened 2022-09-18
GC

angelini:
**This is a big one, we should book some time to go over the technique before reviewing this.**

There are three GC modes added in this PR, but they share most of the same underlying tools.

## Contents GC

In this mode we sample a random set of hashes from the contents table and check if there are any objects that reference them. If not, they are deleted.

This method is useful for finding dangling content that has been leftover by deleting projects. When we delete a project we just delete objects and don't check if this was the last time a piece of content was referenced.

## Project GC

In this mode we specify a specific project and the range of dead objects that should be deleted.

> A "dead" object is an object with a `stop_version` that is NOT null. This means that in the latest version of the filesystem this object is no longer referenced.
> "live" objects have a `stop_version` that is null.

This is range is specified with the `keep` and `from` arguments.

`--keep` is an integer that indicates "how many version since the latest version should we keep around". If we have a project at version 125, we likely only need to keep the history of the last ~100 versions. `--keep 100` would delete dead objects between 0 and 25.

We only need to keep as many versions as we suspect will be queried. If we expect to query "rebuild this project from version 102 to 125", we need to keep version 102 around. However, if we no longer expect to query "rebuild from version 12 to 125" we can drop version 12 and all the versions lower than it.

`--from` is used to prevent the tool from deleting version between 0 and `from`. This will be useful for keeping the template version around or for use with production environments in the future.

Once we've dropped the dead object rows, we scan each of their hashes using the content GC method to drop any content that is no longer referenced.

## Random Projects GC

This is just like project GC, but we sample our project list and run the GC against each project. This mode uses the same `--from` and `--keep` bounds.

---

### Indexes

When we scan the projects table to see if content hashes are dangling, this does not currently hit an index. The objects table can get very large in production and this full table scan will be very slow.

That being said, we should weigh that against the cost of adding another index to the objects table.

Can we tolerate one slow sequential scan every time we GC or should we add an index just for this operation?

### Javascript

This PR does not add these tools to the Javascript client. This remains an open TODO.

---

/cc @airhorns No need to code review this, but you may be interested in the GC technique and index discussion.


internal/db/gc.go
```diff
@@ -0,0 +1,98 @@
...
+		  AND stop_version IS NOT NULL
+		  AND stop_version <= ((SELECT version FROM latest) - $3)
+		RETURNING path, (hash).h1, (hash).h2
+	`, project, fromVersion, keep)
+	if err != nil {
+		return hashes, fmt.Errorf("GcProjectObjects query, project %v, keep %v, from %v: %w", project, keep, fromVersion, err)
+	}
+
+	for rows.Next() {
+		var path string
```
scott-rc: We're not using this `path` -- Can we 🔥 it?

pkg/api/fs.go
```diff
@@ -645,3 +645,148 @@ func (f *Fs) Reset(ctx context.Context, req *pb.ResetRequest) (*pb.ResetResponse
 
 	return &pb.ResetResponse{}, nil
 }
+
+func (f *Fs) GcProject(ctx context.Context, req *pb.GcProjectRequest) (*pb.GcProjectResponse, error) {
+	ctx, span := telemetry.Start(ctx, "fs.gc-project")
+	defer span.End()
```
scott-rc: I've started using the span that `otelgrpc` creates for us instead of starting our own. 

Can we change the beginning of this function and the ones below to something like:

```go
trace.SpanFromContext(ctx).SetAttributes(
	key.Project.Attribute(req.Project),
)
```

pkg/client/client.go
```diff
@@ -557,6 +557,64 @@ func (c *Client) Reset(ctx context.Context, state string) error {
...
+	response, err := c.fs.GcProject(ctx, request)
+	if err != nil {
+		return 0, fmt.Errorf("gc project %v: %w", project, err)
+	}
+
+	return response.Count, nil
+}
+
+func (c *Client) GcRandomProjects(ctx context.Context, sample float32, keep int64, from *int64) (int64, error) {
+	ctx, span := telemetry.Start(ctx, "client.gc-random-project", trace.WithAttributes(
```
scott-rc, line 581: super nit

```suggestion
	ctx, span := telemetry.Start(ctx, "client.gc-random-projects", trace.WithAttributes(
```

pkg/cli/gc.go
```diff
@@ -0,0 +1,86 @@
...
+				}
+				if *from == -1 {
+					from = nil
+				}
+
+				count, err = c.GcProject(ctx, project, keep, from)
+				if err != nil {
+					return fmt.Errorf("could not gc project %v: %w", project, err)
+				}
+			case "random-project":
```
scott-rc: Can we make this plural so that it's clear it's going to gc multiple projects?

```suggestion
			case "random-projects":
```

internal/db/project.go
```diff
@@ -38,3 +38,37 @@ func DeleteProject(ctx context.Context, tx pgx.Tx, project int64) error {
 
 	return nil
 }
+
+func RandomProjects(ctx context.Context, tx pgx.Tx, sample float32) ([]int64, error) {
+	var projects []int64
+
+	for i := 0; i < 5; i++ {
+		// The SYSTEM sampling method would be quicker but it often produces no or all data
+		// on tables with just a few rows.
```
thegedge, line 47: :+1:

internal/db/gc.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+	}
+
+	query := fmt.Sprintf(`
+		WITH missing AS (
+			SELECT c.hash
+			FROM dl.contents c
+			LEFT JOIN dl.objects o
+				ON c.hash = o.hash
+			WHERE c.hash IN (%s)
```
scott-rc: We talked about this IRL but just to document our thoughts.

It would be nice if we used an array parameter here instead of a list of dynamic ones. Otherwise, PGAnalyze will think all of these queries are unique and we won't get nice metrics.

internal/db/gc.go
```diff
@@ -0,0 +1,98 @@
...
+	return tag.RowsAffected(), nil
+}
+
+func encodeHashes(hashes []Hash) []any {
+	fields := make([]any, len(hashes))
+
+	for idx, hash := range hashes {
+		fields[idx] = pgtype.CompositeFields{
+			pgtype.UUID{Bytes: *(*[16]byte)(hash.H1), Valid: true},
+			pgtype.UUID{Bytes: *(*[16]byte)(hash.H2), Valid: true},
```
scott-rc: funky 😵‍💫

scott-rc (approved):
Couple small nits, but otherwise looks great! 👍 🚀

> Can we tolerate one slow sequential scan every time we GC or should we add an index just for this operation?

I'm down to ship this approach and change things up if we run into issues.

pkg/api/fs.go
```diff
@@ -645,3 +645,148 @@ func (f *Fs) Reset(ctx context.Context, req *pb.ResetRequest) (*pb.ResetResponse
...
+
+	tx, close, err := f.DbConn.Connect(ctx)
+	if err != nil {
+		return nil, status.Errorf(codes.Unavailable, "FS db connection unavailable: %v", err)
+	}
+	defer close(ctx)
+
+	if req.KeepVersions <= 0 {
+		return nil, status.Error(codes.InvalidArgument, "Invalid GC KeepVersions: cannot keep 0 versions")
+	}
```
thegedge: nit: Maybe move this above the `DbConn.Connect` to avoid unnecessarily opening a transaction? If gucci, can do the same in `GcRandomProjects`.

pkg/api/fs.go
```diff
@@ -645,3 +645,148 @@ func (f *Fs) Reset(ctx context.Context, req *pb.ResetRequest) (*pb.ResetResponse
...
+
+	projects, err := db.RandomProjects(ctx, tx, req.Sample)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "FS gc random projects %f: %v", req.Sample, err)
+	}
+
+	var allHashes []db.Hash
+
+	for _, project := range projects {
+		hashes, err := db.GcProjectObjects(ctx, tx, project, req.KeepVersions, fromVersion)
```
thegedge: Any reason to do this over giving the full list of projects to `GcProjectObjects` and doing an `project IN (projects)` query?

test/gc_test.go
```diff
@@ -0,0 +1,210 @@
...
+	fs := tc.FsApi()
+
+	response, err := fs.GcProject(tc.Context(), &pb.GcProjectRequest{
+		Project:      1,
+		KeepVersions: 1,
+	})
+	require.NoError(t, err, "fs.GcProject")
+
+	assert.Equal(t, int64(1), response.Project, "Gc result project")
+	assert.Equal(t, int64(2), response.Count, "Gc result count")
```
thegedge, line 34: Just to make sure I'm understanding the tests correctly: the two objects being GC-ed here are

```
writeObject(tc, 1, 1, i(2), "/a", "a v1")
writeObject(tc, 1, 2, i(3), "/b", "b v2")
```

But because we keep a single version, versions 3 and 4 are kept, so we keep this one:

```
writeObject(tc, 1, 3, i(4), "/b", "b v3")
```

thegedge (approved):
👏 

Given we have to trigger this, I think we can focus on our own projects first and ensure correctness/stability in production.

pkg/api/fs.go
angelini: No reason, that would be a better API

test/gc_test.go
angelini: Yeah that's correct.

Scott had the same question though, we technically keep 2 versions here because we keep things that ended in v4.

I don't know that it's strictly necessary to do so, but keeping an extra version also isn't super problematic. Let me see if I can change that.

internal/db/gc.go
angelini: Supposedly Go 1.18 supports slice to byte array conversion `([16]byte)hash.H1` but I couldn't get it to work without this weird pointer indirection...

pkg/api/fs.go
angelini: Upon further review the query depends on the project's latest version - which is a bit tricky to write for multiple projects in one SQL query.

Instead I'll create a multiple projects helper that does this loop within the `db.` package

pkg/server/server.go
```diff
@@ -55,8 +57,26 @@ func NewDbPoolConnector(ctx context.Context, uri string) (*DbPoolConnector, erro
...
+	defer conn.Release()
+
+	var extraTypes []*pgtype.Type
+	for _, typeName := range []string{"hash", "hash[]"} {
+		extraType, err := conn.Conn().LoadType(ctx, typeName)
+		if err != nil {
+			return nil, fmt.Errorf("could not load type %s: %w", typeName, err)
+		}
+
+		conn.Conn().TypeMap().RegisterType(extraType)
```
scott-rc, line 73: Do we need to do this here if we always do it in `Connect` below?

angelini: We do need this because you cannot register the `hash[]` type if you haven't first registered the `hash` type.

And then we do it on every `Connect` because a new connection from the pool may not have had these types registered. `LoadType` is slow (hits the db), `RegisterType` is very quick (assigns a value)

PR #4, opened 2022-09-26
[node-module-cache] Introduce cache computation for impactful node_modules

gmalette:
This PR is the first in a series to add caches for `node_modules` with the Greatest Impact ™️ 

Specifically, this PR adds:
- A new `cache_versions` table
- The ability to compute a new version of cached modules (although it currently can't be called from the outside)
- A gRPC call to get the cache as tarballs ~ (`func GetCache() returns (stream CachedObjectTar)`)

## Caching Approach, Explained

The way we'll provide caching from an `app-sandbox` K8s pod's perspective, is that DL will fetch the cache and store it in a folder. DL will know which `cache_version` are available, and provide them as arguments when calling `GetCompress`. The returned objects will be signaled if they're available as part of the given `cache_version`, and their payload will not be returned. 

One caveat of this approach is that `cache_version` must remain absolutely coherent and be immutable. 

## `cache_versions` table

The `cache_versions` table consists of two columns:
- `version bigserial PRIMARY KEY`
- `hashes hash[]`

This will produce a single row per version.

An alternative could've been `version bigint, hash hash` and use one row per hash, and multiple rows per version. 

I chose the one-row-per-version solution for one main reason: it's easier to show it's consistent, harder to make mistakes.

With the multiple-rows-per-version approach, it's easier to add rows to a version, either through race conditions or by accident. This would lead to clients breaking.

With one-row-per-version, it's much harder to accidentally change it given the application will likely never perform `UPDATE` statements on the `cache_versions` table. If we wanted to make the table more robust, we could even remove `UPDATE` from the permissions granted.

## Greatest Impact ™️ Definition

Right now we're a bit naive in selecting the packages to cache, but in all likelihood, it will nonetheless perform quite well.

What we're doing is selecting `node_modules` that have the greatest number of still-used (no `stop_version`) occurrences, regardless of size or deployment counts.


pkg/api/fs.go
```diff
@@ -645,3 +645,46 @@ func (f *Fs) Reset(ctx context.Context, req *pb.ResetRequest) (*pb.ResetResponse
...
+	}
+
+	for {
+		version, tar, packPath, err := tars()
+		if err == io.EOF {
+			break
+		}
+		// if err == db.SKIP {
+		//	continue
+		//}
```
gmalette, line 673: As far as I understand, `SKIP` is only returned when the results are filtered, so I don't think we need to handle it, do we?

internal/db/query.go
```diff
@@ -274,6 +275,58 @@ func GetTars(ctx context.Context, tx pgx.Tx, project int64, vrange VersionRange,
...
+
+	stream := func() (int64, []byte, *string, error) {
+		if !rows.Next() {
+			return 0, nil, nil, io.EOF
+		}
+		var rowVersion int64
+		var hash string
+		var encoded []byte
+
+		err := rows.Scan(&rowVersion, &hash, &encoded)
```
angelini, line 314: If you pull the latest from main we now have a `Hash` type for these instead of just `[]byte`.

And to read it easily from PGX (there are a few examples hanging around) you can do something like:

```
SELECT (hash).h1, (hash).h2
```

```
var hash Hash
rows.Scan(&hash.H1, &hash.H2)
```

internal/db/query.go
```diff
@@ -315,3 +368,32 @@ func (p *PackManager) IsPathPacked(path string) *string {
...
+	sql := `
+		WITH impactful_packed_objects AS (
+			SELECT hash, count(*) as count
+			FROM dl.objects
+			WHERE path LIKE $1
+				AND packed = true
+				AND stop_version IS NULL
+			GROUP BY hash
+			ORDER BY count DESC
+			LIMIT 100
```
angelini, line 383: Should we make this limit value an argument to this function?

Feels like the kind of thing we'll be tuning a lot and may even make an arg to the create cache RPC.

internal/pb/fs.proto
```diff
@@ -126,3 +128,16 @@ message ResetRequest {
 }
 
 message ResetResponse {};
+
+message GetCacheRequest {};
+
+message GetCacheResponse {
+    enum Format {
```
angelini, line 135: I'm not super familiar with the proto format, but I suspect we can declare this outside of the `message` and then share the type between `GetCacheResponse` and `GetCompressResponse`.

pkg/api/fs.go
angelini: Yeah that's right, `SKIP` was something I added and not part of the typical responses. Not needed here.

pkg/api/fs.go
```diff
@@ -645,3 +645,46 @@ func (f *Fs) Reset(ctx context.Context, req *pb.ResetRequest) (*pb.ResetResponse
...
+		if err != nil {
+			return status.Errorf(codes.Internal, "FS get next tar: %v", err)
+		}
+
+		err = stream.Send(&pb.GetCacheResponse{
+			Version:  version,
+			Format:   pb.GetCacheResponse_S2_TAR,
+			Bytes:    tar,
+			PackPath: *packPath,
```
angelini, line 682: Why do they end up needing this path? Is the path information not already contained within the Tar file?

gmalette:
Closed in favour of #8

PR #5, opened 2022-09-26
Clone to project

jasong689:
This involved a few more changes than I had anticipated. This PR adds a `CloneToProject` method to copy changes from one project to another. One thing to note is that the way it's currently implemented is as a one way copy where only changes from the source (from a `fromVersion` up to a `toVersion`) are copied over to the target project. Any objects that exist in the target which haven't explicitly been removed from the source are not removed. This will be used for development to production deploys where we copy changes since the last deploy to the production project.

closes GGT-1576


test/shared_test.go
```diff
@@ -392,6 +392,8 @@ func debugObjects(tc util.TestCtx) {
 	rows, err := conn.Query(tc.Context(), `
 		SELECT project, start_version, stop_version, path, mode, size, packed, (hash).h1, (hash).h2
 		FROM dl.objects
+		ORDER BY
+			project, start_version, stop_version, path
```
jasong689: It was much easier to view the debug data once it had been sorted

internal/pb/fs.proto
```diff
@@ -64,6 +66,7 @@ message ObjectQuery {
     bool is_prefix = 2;
     bool with_content = 3;
     repeated string ignores = 4;
+    optional bool with_hash = 5;
```
jasong689: I went with an optional here. Other option I think is to make this required and handle current calls with a default value of `false` in the client?

pkg/api/fs.go
```diff
@@ -101,6 +101,63 @@ func (f *Fs) NewProject(ctx context.Context, req *pb.NewProjectRequest) (*pb.New
 	return &pb.NewProjectResponse{}, nil
 }
 
+func (f *Fs) CloneToProject(ctx context.Context, req *pb.CloneToProjectRequest) (*pb.CloneToProjectResponse, error) {
+	trace.SpanFromContext(ctx).SetAttributes(
+		key.Project.Attribute(req.Source),
+		key.FromVersion.Attribute(&req.FromVersion),
```
jasong689, line 107: These are reused from other methods, not 100% sure if that's appropriate

internal/db/queryBuilder.go
```diff
@@ -136,25 +148,38 @@ func (qb *queryBuilder) queryWithRemovals() string {
 	%s
 	)
 
-	SELECT path, mode, size, bytes, packed, deleted
+	SELECT path, mode, size, bytes, packed, deleted, hash
```
jasong689: The hash needs to be returned so that it can be copied into the `target` project

internal/db/copy.go
```diff
@@ -42,3 +40,73 @@ func CopyAllObjects(ctx context.Context, tx pgx.Tx, source int64, target int64)
...
 }
+
+func CloneToProject(ctx context.Context, tx pgx.Tx, source int64, target int64, vrange VersionRange) (*int64, error) {
+	withHash := true
+
+	objectQuery := &pb.ObjectQuery{
+		Path:        "",
+		IsPrefix:    true,
+		WithContent: false,
+		WithHash:    &withHash,
```
jasong689: Is there a better approach than to add this to protobuf object (it changes the grpc signatures)?

internal/db/queryBuilder.go
```diff
@@ -136,25 +148,38 @@ func (qb *queryBuilder) queryWithRemovals() string {
...
 	FROM updated_objects
-	UNION
-	SELECT path, mode, size, bytes, packed, deleted
+	UNION ALL
```
jasong689: I changed this since `deleted` is always different between `updated_objects` and `removed_objects`

Makefile
```diff
@@ -32,6 +32,8 @@ install:
 	go install -tags 'postgres' github.com/golang-migrate/migrate/v4/cmd/migrate@v4.14
 	go install github.com/bojand/ghz/cmd/ghz@v0.105.0
 	go install github.com/gadget-inc/fsdiff/cmd/fsdiff@v0.4
+	go install -v github.com/stamblerre/gocode@latest
```
jasong689: These were required by VSCode golang tooling. Figured I'd add them here

jasong689:
Would like to add some more test scenarios, but figured I'd tag you (@angelini and @scott-rc) to get the review started

pkg/api/fs.go
scott-rc: The `FromVersion` and `ToVersion` makes sense to be reused, however, I'm not sure we should use `Project` for `req.Source` and `CloneToProject` for `req.Target`. I prefer to match keys to their variables whenever possible.

How do we feel about renaming `source`/`target` to `from_project`/`to_project` and then adding `FromProject`/`ToProject` to the `keys` package?

pkg/client/client.go
```diff
@@ -557,6 +557,29 @@ func (c *Client) Reset(ctx context.Context, state string) error {
 	return nil
 }
 
+func (c *Client) CloneToProject(ctx context.Context, source int64, target int64, fromVersion int64, toVersion int64) (*int64, error) {
+	ctx, span := telemetry.Start(ctx, "client.cloneToProject", trace.WithAttributes(
```
scott-rc:
```suggestion
	ctx, span := telemetry.Start(ctx, "client.clone-to-project", trace.WithAttributes(
```

internal/pb/fs.proto
scott-rc: I'm pretty sure removing `optional` will work as intended. According to [the docs](https://developers.google.com/protocol-buffers/docs/proto3#default), missing properties are automatically set to their default value (`false` for booleans).

internal/db/copy.go
scott-rc: I think you could add `withHash` to the `queryBuilder` struct and pass the value to the `newQueryBuilder` function instead.

jasong689: Ah yea that could work, but knowing that protobuf defaults to false for missing properties maybe keeping it here is ok.

pkg/api/fs.go
jasong689: I actually had these originally named `FromProject` to `ToProject` but kept getting them confused with `FromVersion` and `ToVersion`. I'm down to change them back though

scott-rc: Oh interesting! I thought using `FromProject`/`ToProject` would be nice because it matches our `ToVersion`/`FromVersion`, but if making them similar causes more confusion than clarity then I'm down to just add `Source` and `Target` to the `keys` package.

I'll let you decide 👍

Makefile
angelini: Makes sense, can you just drop the `-v` so it lines up with the other installs

internal/db/copy.go
```diff
@@ -4,15 +4,13 @@ import (
...
 	"github.com/jackc/pgx/v5"
 )
 
 func CopyAllObjects(ctx context.Context, tx pgx.Tx, source int64, target int64) error {
-	var samePackPatterns bool
-	err := tx.QueryRow(ctx, `
-		SELECT COALESCE((SELECT pack_patterns FROM dl.projects WHERE id = $1), '{}') =
-		       COALESCE((SELECT pack_patterns FROM dl.projects WHERE id = $2), '{}');
-	`, source, target).Scan(&samePackPatterns)
+	samePackPatterns, err := HasSamePackPattern(ctx, tx, source, target)
```
angelini, line 12: Makes sense to pull this out into it's own function, but I don't think we need it to be exported, it can remain private to this package

internal/db/copy.go
```diff
@@ -42,3 +40,71 @@ func CopyAllObjects(ctx context.Context, tx pgx.Tx, source int64, target int64)
...
+	objectQuery := &pb.ObjectQuery{
+		Path:        "",
+		IsPrefix:    true,
+		WithContent: false,
+		WithHash:    true,
+	}
+
+	latestVersion, err := LockLatestVersion(ctx, tx, target)
+
+	if err != nil {
```
angelini: Nit: The pattern in this codebase is to keep the error checks right under the line where the error is defined, instead of including an extra blank line.

internal/db/copy.go
```diff
@@ -42,3 +40,71 @@ func CopyAllObjects(ctx context.Context, tx pgx.Tx, source int64, target int64)
...
+	removedArgsLength := len(removedArgs)
+
+	sql := fmt.Sprintf(`
+	%v
+	UPDATE dl.objects
+	SET stop_version = $%d
+	FROM removed_objects
+	WHERE "removed_objects".path = "objects".path
+	AND "objects".project = $%d
+	AND "objects".stop_version IS NULL
```
angelini: Nit: Also the pattern in this codebase is to ident SQL one more time. And look for examples of how `WHERE` & `AND` are formatted elsewhere in the codebase.

```
WHERE x = 1
  AND y = 2
```

internal/db/copy.go
```diff
@@ -42,3 +40,71 @@ func CopyAllObjects(ctx context.Context, tx pgx.Tx, source int64, target int64)
...
+	}
+
+	newTargetVersion := latestVersion + 1
+
+	builder := newQueryBuilder(source, vrange, objectQuery)
+	removedSql, removedArgs := builder.buildRemoved()
+	removedArgsLength := len(removedArgs)
+
+	sql := fmt.Sprintf(`
+	%v
```
angelini: We should use a string pattern `%s` not the debugging one.

internal/db/copy.go
```diff
@@ -42,3 +40,71 @@ func CopyAllObjects(ctx context.Context, tx pgx.Tx, source int64, target int64)
...
+
+func CloneToProject(ctx context.Context, tx pgx.Tx, source int64, target int64, vrange VersionRange) (*int64, error) {
+	objectQuery := &pb.ObjectQuery{
+		Path:        "",
+		IsPrefix:    true,
+		WithContent: false,
+		WithHash:    true,
+	}
+
+	latestVersion, err := LockLatestVersion(ctx, tx, target)
```
angelini: This lock should be taken in `fs.go` as the other GRPC methods do.

internal/pb/fs.proto
```diff
@@ -70,6 +72,7 @@ message ObjectQuery {
     bool is_prefix = 2;
     bool with_content = 3;
     repeated string ignores = 4;
+    bool with_hash = 5;
```
angelini: I don't think we should add `with_hash` here.

This is the public facing API for DateiLager and `with_hash` is only ever used internally. Instead we should just make this a function argument to the query builder and not something we expose.

internal/db/queryBuilder.go
```diff
@@ -185,3 +208,27 @@ func (qb *queryBuilder) build() (string, []any) {
...
+	}
+
+	query, args := qb.replaceQueryArgs(query)
+
+	return query, args
+}
+
+func AsCte(query string, alias string) string {
+	template := `
+	WITH %s as (
```
angelini: As mentioned elsewhere, SQL is idented an extra time and `as` is a SQL identifier and should be capitalized.

angelini:
It looks like these changes will have the `queryBuilder` generated nested CTEs.

I don't think that's really necessary considering what these queries look like, we should be able to produce flattened queries with only one level of CTEs.

These queries are already tricky to parse for humans, let's no make it any harder - and they are often reviewed and debugged in PgAnalyze, we should make them as uniform and easy to parse as possible.

internal/db/copy.go
```diff
@@ -42,3 +40,71 @@ func CopyAllObjects(ctx context.Context, tx pgx.Tx, source int64, target int64)
...
+
+	changedArgs = append(changedArgs, target, newTargetVersion)
+
+	_, err = tx.Exec(ctx, sql, changedArgs...)
+
+	if err != nil {
+		return nil, fmt.Errorf("copy to project could not insert updated files: %w", err)
+	}
+
+	err = UpdateLatestVersion(ctx, tx, target, newTargetVersion)
```
angelini: This should also be in the `fs.go` RPC handler.

The goal is to try and avoid mixing raw SQL and the steps necessary to process an RPC.

internal/db/queryBuilder.go
```diff
@@ -81,25 +81,24 @@ func (qb *queryBuilder) updatedObjectsCTE() string {
...
+	hashSelector := "null::hash as hash"
+	if qb.objectQuery.WithHash {
+		hashSelector = "o.hash"
+	}
+
+	return fmt.Sprintf(template, bytesSelector, hashSelector, contentsPredicate, pathPredicate, ignoresPredicate)
 }
 
-func (qb *queryBuilder) removedObjectsCTE() string {
+func (qb *queryBuilder) removedObjectsCTE(skipChanged bool) string {
```
angelini: Why would we ever want to generate the `removed_objects` CTE without including changed files. I don't think there's a case where that is valid.

pkg/api/fs.go
```diff
@@ -101,6 +101,63 @@ func (f *Fs) NewProject(ctx context.Context, req *pb.NewProjectRequest) (*pb.New
...
+	vrange, err := db.NewVersionRange(ctx, tx, req.Source, &req.FromVersion, &req.ToVersion)
+
+	if errors.Is(err, db.ErrNotFound) {
+		return nil, status.Errorf(codes.NotFound, "FS get missing latest version: %v", err)
+	}
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "FS get latest version: %v", err)
+	}
+
+	newVersion, err := db.CloneToProject(ctx, tx, req.Source, req.Target, vrange)
```
angelini: We should have a debug log line that include the source, target and vrange

test/shared_test.go
angelini: Sounds good, but the formatting pattern is to keep the ORDER BY args on the same line as the ORDER BY (just as like SELECT)

js/src/grpc-client.ts
```diff
@@ -317,6 +319,34 @@ export class DateiLagerGrpcClient {
...
   }
+
+  /**
+   * Clones the `source`projects changes (from `fromVersion` up to `toVersion`) to the `target` project. This method assumes that it is always a one way clone from source to target, it does not take into account the changes that have occurred in the `target` project.
+   *
+   * @param source      The source project.
+   * @param target      The target project.
+   * @param fromVersion Start version of the source project.
+   * @param toVersion   Stop version of the source project.
+   * @returns             The new version number of hte target project
```
angelini: Nit: the indentation is off here.

internal/db/queryBuilder.go
jasong689: So I changed this because I wanted to be able to delete files which have changed and replace them with their new versions. This is 2 steps where we set a stop version on the removed files and then an insert on all changed files. If we filtered out changed files we would have to intersect the existing files in target with changed files in source

internal/db/queryBuilder.go
```diff
@@ -111,50 +118,66 @@ func (qb *queryBuilder) removedObjectsCTE() string {
...
+func (qb *queryBuilder) queryWithoutRemovals(asCte bool) string {
 	template := `
-	WITH possible_objects AS (
-	%s
-	), updated_objects AS (
-	%s
-	), removed_objects AS (
-	%s
-	)
-
```
jasong689, line 138: I removed the space between the CTE and select statement. Was that important for pganalyze stats?

internal/db/queryBuilder.go
```diff
@@ -185,3 +208,17 @@ func (qb *queryBuilder) build() (string, []any) {
 
 	return query, args
 }
+
+func (qb *queryBuilder) build(asCte bool) (string, []any) {
```
jasong689: Didn't think a separate method was necessary so I added an arg to return the query as a CTE instead of a select statement

js/src/grpc-client.ts
```diff
@@ -317,6 +317,34 @@ export class DateiLagerGrpcClient {
...
   }
+
+  /**
+   * Clones the `source`projects changes (from `fromVersion` up to `toVersion`) to the `target` project. This method assumes that it is always a one way clone from source to target, it does not take into account the changes that have occurred in the `target` project.
+   *
+   * @param source      The source project.
+   * @param target      The target project.
+   * @param fromVersion Start version of the source project.
+   * @param toVersion   Stop version of the source project.
+   * @returns           The new version number of hte target project
```
scott-rc:
```suggestion
   * Clones the `source` project's changes (from `fromVersion` up to `toVersion`) to the `target` project. This method assumes that it is always a one way clone from source to target, it does not take into account the changes that have occurred in the `target` project.
   *
   * @param source      The source project.
   * @param target      The target project.
   * @param fromVersion Start version of the source project.
   * @param toVersion   Stop version of the source project.
   * @returns           The new version number of the target project
```

jasong689: English is hard 😄

angelini (approved):

scott-rc (approved):

PR #6, opened 2022-09-27
Recover from arbitrary errors writing files by removing the thing we're about to write

airhorns:
I encountered a sequence of changes that DL didn't expect where I was creating a symlink at a path that was previously a folder full of files. There's a bunch of other annoying cases like this, where we should be able to go from any state on disk to any new incoming object type.

 The existing strategy did a bit of checking to see if stuff wasn't quite right, but I think we aught to be a bit more robust. The recovery strategy is always the same if we have a new incoming object: delete whatever is there so we can just write the new thing on top. If we're gonna do that, I think we can do it a bit more blindly. 

Instead of checking ahead of time for erroneous conditions (which also requires another syscall every write), we just blindly try to do the thing we need to do, and if it fails, remove whatever is there and try again. That way, we don't need to anticipate whatever fucked up case is on disk at the time and instead just blow it away. I also like this because it makes the fast path a bit faster where we do less checking up front and just let the kernel tell us we're doing something weird.

### DIFF
--- a/internal/files/writer.go
+++ b/internal/files/writer.go
@@ -17,32 +17,20 @@ func fileExists(path string) bool {
 	return err == nil
 }
 
-func removePathIfSymlink(path string) error {
-	stat, err := os.Lstat(path)
+type retryableFileOperation[R any] func() (R, error)
+
+// if we encounter an error making a directory, symlink, or opening a file, we try to recover by just removing whatever is currently at the path, and trying again
+func retryFileErrors[R any](path string, fn retryableFileOperation[R]) (R, error) {
+	result, err := fn()
+
 	if err != nil {
-		return nil
+		err = os.RemoveAll(path)
+		if err != nil {
+			return result, fmt.Errorf("removing existing path error %v: %w", path, err)
+		}
+		return fn()
 	}
-
-	if stat.Mode()&os.ModeSymlink == os.ModeSymlink {
-		err = os.Remove(path)
-		return err
-	}
-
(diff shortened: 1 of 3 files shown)


internal/files/writer.go
```diff
@@ -72,33 +58,32 @@ func writeObject(rootDir string, reader *db.TarReader, header *tar.Header) error
 		if err != nil {
 			return fmt.Errorf("write %v to disk: %w", path, err)
 		}
-
-	case tar.TypeDir:
-		err := removePathIfNotDirectory(path)
+		err = os.Chmod(path, os.FileMode(header.Mode))
```
angelini, line 61: Is this `Chmod` necessary if we just opened the file with the correct mode?

angelini (approved):
This seems like a much easier path to getting this right once and for all.

This file was littered with my attempts to catch every case - and I clearly forgot quite a few still.

🚢 🇮🇹

internal/files/writer.go
airhorns: It isn't if we just created the file, but if it was already there and has the wrong mode, opening it doesn't change the mode, so we need to do this explicitly after. Kind of unfortunate cause it is another syscall. I couldn't figure out a way to get the open call to tell us if it created the file or just opened an existing one, cause we only need to do this for things we didn't create. Also if there's a function which opens-and-returns-a-info object that'd work too.

scott-rc (approved):

PR #8, opened 2022-09-30
DL Caching on the client

gmalette:
This PR adds caching to DL both on the server side and on the client side. The branch itself is based on #4 (server-side), in case we want to merge server and client separarely. I would strongly encourage reading #4 for a description of the changes for `fs.GetCache()`.

## `client.GetCache()`

The client has a new `GetCache()` function which allows populating the provided cache directory with the contents returned by the server. The hierarchy changed from the original proposal:

```
<cache root> /
    - .lock
    - versions
    - tmp/
    - objects/
        <cache key> /
            <packed content>
```

The `.lock` file is only present when an ongoing `GetCache()` request is performed on this cache root. The goal is to avoid two processes racing to populate the same cache root. The file is created with `O_EXCL`, meaning that only one process can create this file; others will get an error and abort.

The `versions` file is a plain-text file containing the versions available.

The `tmp` directory is where the files are extracted. This is done to prevent an extraction failure from corrupting the cached objects. The files are then moved to their final destination, in the `objects` directory.

**NOTE**: If the process crashes while fetching the cache, the `.lock` file will not be cleaned up and it will not be possible for other processes to populate the cache root. This was chosen in order to avoid undefined behaviour within the cache.

## `fs.GetCompress` 

This PR mainly changes `fs.GetCompress` and its dependencies to allow caching packages. 

- A new CTE is added to the `updated_objects`, allowing a `LEFT JOIN` on the cached versions sent by the client
- The performed DB queries will no longer return the `bytes` column for a cached entry
- I'm abusing the tarfile content to send the cache hash. On the client-side the hash is hex-encoded to produce the path in the cache folder, to which we append the file name and symlink there.


internal/db/query.go
```diff
@@ -165,12 +166,12 @@ func GetObjects(ctx context.Context, tx pgx.Tx, packManager *PackManager, projec
...
 	}
 
-	builder := newQueryBuilder(project, vrange, objectQuery, false)
+	builder := newQueryBuilder(project, vrange, objectQuery, nil, false)
 	sql, args := builder.build()
 	rows, err := tx.Query(ctx, sql, args...)
 
 	if err != nil {
-		return nil, fmt.Errorf("getObjects query, project %v vrange %v: %w", project, vrange, err)
+		return nil, fmt.Errorf("1getObjects query, project %v vrange %v: %w", project, vrange, err)
```
angelini: I think this change was a typo.

internal/db/query.go
```diff
@@ -189,16 +190,21 @@ func GetObjects(ctx context.Context, tx pgx.Tx, packManager *PackManager, projec
...
 		var path string
 		var mode, size int64
+		var cache_h1, cache_h2 []byte
 		var encoded []byte
 		var packed bool
 		var deleted bool
 		var h1, h2 []byte
 
-		err := rows.Scan(&path, &mode, &size, &encoded, &packed, &deleted, &h1, &h2)
+		err := rows.Scan(&path, &mode, &size, &cache_h1, &cache_h2, &encoded, &packed, &deleted, &h1, &h2)
```
angelini: As mentioned in the other PR, we can use the `Hash` type here.

```
var hash Hash
rows.Scan(&hash.H1, &hash.H2)
```

internal/db/query.go
```diff
@@ -250,31 +256,39 @@ func GetTars(ctx context.Context, tx pgx.Tx, project int64, vrange VersionRange,
...
 		if err != nil {
 			return nil, nil, fmt.Errorf("getTars scan, project %v vrange %v: %w", project, vrange, err)
 		}
 
-		if packed {
-			return encoded, &path, nil
+		if cache_h1 != nil {
+			cache_hash = make([]byte, 32)
+			copy(cache_hash[0:16], cache_h1)
+			copy(cache_hash[16:], cache_h2)
```
angelini: This can be replaced by the `.Bytes()` method on `Hash`

internal/db/query.go
```diff
@@ -332,3 +402,32 @@ func (p *PackManager) IsPathPacked(path string) *string {
...
+	sql := `
+		WITH impactful_packed_objects AS (
+			SELECT hash, count(*) as count
+			FROM dl.objects
+			WHERE path LIKE $1
+				AND packed = true
+				AND stop_version IS NULL
+			GROUP BY hash
+			ORDER BY count DESC
+			LIMIT 100
```
angelini: As mentioned in the other PR, should this become a function var and maybe even an argument to the `CreateCache` RPC.

internal/db/queryBuilder.go
```diff
@@ -118,29 +122,39 @@ func (qb *queryBuilder) removedObjectsCTE() string {
...
+}
+
 func (qb *queryBuilder) queryWithoutRemovals() string {
 	template := `
 		WITH possible_objects AS (
 		%s
-		), updated_objects AS (
+		), cached_object_hashes AS (
+	%s
+	), updated_objects AS (
```
angelini: Nit: There's some formatting misalignment here that shows up in Github.

internal/db/queryBuilder.go
```diff
@@ -149,18 +163,18 @@ func (qb *queryBuilder) queryWithRemovals() string {
 	`
 
 	selectStatement := `
-		SELECT path, mode, size, bytes, packed, deleted, (hash).h1, (hash).h2
+		SELECT path, mode, size, (cache_hash).h1 cache_h1, (cache_hash).h2 cache_h2, bytes, packed, deleted, (hash).h1, (hash).h2
```
angelini: Is the `cache_hash` different than the `hash`?

It seems like we may be returning the same values twice. Could we always return `hash` and then include a `is_hashed` boolean?

internal/files/writer.go
```diff
@@ -2,13 +2,16 @@ package files
...
 	"archive/tar"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
 	"io/fs"
 	"os"
 	"path/filepath"
 
+	"github.com/gadget-inc/dateilager/internal/pb"
```
angelini: Is this importing the same module twice?

internal/pb/fs.go
```diff
@@ -9,7 +9,12 @@ import (
 	"path/filepath"
 )
 
-func tarTypeFromMode(mode fs.FileMode) byte {
+const (
+	TarDeleted = 'D'
+	TarCached  = 'C'
```
angelini, line 14: Sounds like we're good to use `'C'`

image

pkg/api/fs.go
```diff
@@ -862,3 +862,46 @@ func (f *Fs) GcContents(ctx context.Context, req *pb.GcContentsRequest) (*pb.GcC
...
+	if err != nil {
+		return status.Errorf(codes.Internal, "FS get cached tars: %v", err)
+	}
+
+	for {
+		version, tar, hash, err := tars()
+		if err == io.EOF {
+			break
+		}
+		// if err == db.SKIP {
```
angelini: As mention in the other PR, we don't need to handle this here since we always get back then entire cache.

test/db_test.go
```diff
@@ -0,0 +1,156 @@
...
+	var hashes [][]byte
+
+	for rows.Next() {
+		var h1, h2 []byte
+		err = rows.Scan(&h1, &h2)
+		require.NoError(t, err)
+
+		hash := make([]byte, 32)
+		copy(hash[0:16], h1)
+		copy(hash[16:32], h2)
```
angelini: This can also make use of the `Hash.Bytes()` method

pkg/client/client.go
```diff
@@ -557,6 +562,177 @@ func (c *Client) Reset(ctx context.Context, state string) error {
...
+	defer versionFile.Close()
+	if _, err = versionFile.WriteString(fmt.Sprintf("%d\n", version)); err != nil {
+		return -1, fmt.Errorf("fs.GetCache failed to update the versions file: %w", err)
+	}
+
+	err = cleanupCacheLockFile(lockFile)
+	if err != nil {
+		return -1, err
+	}
+	os.RemoveAll(tmpObjectDir)
```
angelini: You can set this up as a `defer` right after creating the dir, which will ensure that gets removed even in error cases.

pkg/client/client.go
```diff
@@ -557,6 +562,177 @@ func (c *Client) Reset(ctx context.Context, state string) error {
...
+	versionFile, err := os.OpenFile(cacheVersionPath(cacheRootDir), os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0600)
+	if err != nil {
+		return -1, fmt.Errorf("fs.GetCache cannot open cache versions file for writing: %w", err)
+	}
+	defer versionFile.Close()
+	if _, err = versionFile.WriteString(fmt.Sprintf("%d\n", version)); err != nil {
+		return -1, fmt.Errorf("fs.GetCache failed to update the versions file: %w", err)
+	}
+
+	err = cleanupCacheLockFile(lockFile)
```
angelini: So we kind of want to make sure this happens no matter what.

What do you think if we made it so that this function cannot return an error and then add it as a `defer` right after we create the lock.

Trying to avoid cases where we can potentially leave a directory in a bad state that cannot be recovered from.

pkg/client/client.go
```diff
@@ -557,6 +562,177 @@ func (c *Client) Reset(ctx context.Context, state string) error {
...
+				if err != nil {
+					cancel()
+					return fmt.Errorf("fs.GetCache receive: %w", err)
+				}
+			}
+		}
+	})
+
+	for i := 0; i < parallelWorkerCount(); i++ {
+		// create the attribute here when `i` is different
```
angelini, line 661: 🙌 

There's actually a proposal to fix this in upcoming Go versions where the beginning of every loop will have an implicit `i := i`

pkg/client/client.go
```diff
@@ -557,6 +562,177 @@ func (c *Client) Reset(ctx context.Context, state string) error {
...
+					tempDest := filepath.Join(tmpObjectDir, hashHex)
+					finalDest := filepath.Join(objectDir, hashHex)
+
+					if fileExists(finalDest) {
+						return nil
+					}
+
+					if fileExists(tempDest) {
+						err := os.RemoveAll(tempDest)
+						return fmt.Errorf("temporary cache folder exists for %s and couldn't be removed: %w", tempDest, err)
```
angelini: Should this be `if err != nil`?

And should we call `cancel()` in this case?

pkg/client/client.go
```diff
@@ -557,6 +562,177 @@ func (c *Client) Reset(ctx context.Context, state string) error {
...
+		return -1, nil
+	}
+	if err != nil {
+		return -1, fmt.Errorf("fs.GetCache receive: %w", err)
+	}
+	version := response.Version
+
+	for _, availableVersion := range availableVersions {
+		if version == availableVersion {
+			return availableVersions[len(availableVersions)-1], nil
```
angelini: It's not clear to me why we're returning the last `availableVersion` in this case.

pkg/client/client.go
```diff
@@ -557,6 +562,177 @@ func (c *Client) Reset(ctx context.Context, state string) error {
...
+					return ctx.Err()
+				case response, ok := <-tarChan:
+					if !ok {
+						return nil
+					}
+
+					tarReader := db.NewTarReader(response.Bytes)
+					hashHex := hex.EncodeToString(response.Hash)
+					tempDest := filepath.Join(tmpObjectDir, hashHex)
+					finalDest := filepath.Join(objectDir, hashHex)
```
angelini, line 680: Why do we need both the `temp` and `final` destinations? Since we're taking a lock is it not safe to write directly into the final destination?

internal/db/query.go
```diff
@@ -291,6 +305,62 @@ func GetTars(ctx context.Context, tx pgx.Tx, project int64, vrange VersionRange,
...
+		}
+		if version != 0 && rowVersion != version {
+			return 0, nil, nil, fmt.Errorf("GetCacheTars rowVersion mismatch: %d vs %d", version, rowVersion)
+		}
+
+		version = rowVersion
+
+		hash := make([]byte, 32)
+		copy(hash[0:16], h1)
+		copy(hash[16:], h2)
```
angelini: Another case where the `Hash` type will add better type info to this function and allow you to not have to repeat this code.

internal/files/writer.go
gmalette: which other instance are you referring to?

pkg/client/client.go
gmalette: You mean, as opposed to the one returned by the server? Since we're skipping the unpacking, I figure that returning the last one that was received makes sense, but now that I say it out loud perhaps not. 

I'll change it to "the one that the server returned" but TBF I've been struggling with the return valud of `GetCache` and I think it should probably either be `void` or `[]int64`

pkg/client/client.go
gmalette: Well, the thing is that we don't know that the extraction will succeed. If we extract directly to the final destination we risk creating an invalid file structure. At present, even if we wrote a bad file, we would not use it, but I can't guarantee that this will hold, and I haven't written tests to prove it either.

The easiest way to avoid that situation is to extract somewhere else and avoid moving if the extraction fails.

pkg/client/client.go
gmalette: I very much disagree, especially since we currently do not have any recovery mechanisms. 

I'd very much rather have an irrecoverable cache that cannot be used than an inconsistent cache that *may* be recovered from.

In the former case, we're going to be paying more for the transfer of packed files because the cache will be useless (and irrecoverable).

In the latter case, we'd have to deal with multiple, unknown failure modes. We'd have to reason about what all those are before trying to recover from a failure.

Sidenote: I'm starting to think it's harmful to describe the current solution as a "cache", at least from the client's perspective. If there's a cache miss, the application will fail to boot; it will not run in a degraded performance kind of way.

internal/db/queryBuilder.go
gmalette: Yeah possibly. I don't know why `hash` isnt't always returned; it was added during one of the rebases. I'll try to find out.

internal/files/writer.go
angelini: Ah that was a mistake on my part, `internal/pb` and `internal/db` look very similar 🤦 

It looks like you likely don't have the Go imports sorter installed. It should have sorted these alphabetically and removed the newline below here.

pkg/client/client.go
angelini: Ahh got it.

Our two options are to work in a temp directory (as you are doing) or do a full cleanup (`rm -rf`) at the beginning of Cache Creation to potentially drop any failed previous attempt.

Is that second option possible / simpler, or not something we can do in this case.

internal/files/writer.go
gmalette: I think it's because it's not imported in the `.golangci.yml` file. Since I'm importing that file, my editor doesn't let me enable extra ones 🤷🏼‍♂️

pkg/client/client.go
gmalette: A worker could have an in-flight request using existing cache versions. We can *never* delete what has been made available, unless we implement synchronization

angelini: Yeah 100%, we should never delete something once it has been made available.

But if it was never made available then it should be safe to delete?

gmalette: yes, if and only if it was never made available, we can delete it, the trick is in knowing

PR #10, opened 2022-10-05
Add retry policy to our clients

angelini:
Following the docs to create a service config, which is how you introduce a retry policy.

However, I cannot get this to work in any of my testing 😢 

Following almost exactly the example from the GRPC-Go library: https://github.com/grpc/grpc-go/tree/c44f627fd1f65c4e9f2837c17b4a734c516172fd/examples/features/retry

And from the other docs: https://github.com/grpc/grpc/blob/master/doc/service_config.md

But this does not seem to retry in my testing.

### DIFF
--- a/js/src/grpc-client.ts
+++ b/js/src/grpc-client.ts
@@ -77,6 +77,22 @@ export class DateiLagerGrpcClient {
   public constructor(options: DateiLagerGrpcClientOptions) {
     const tokenFn = typeof options.token === "string" ? () => Promise.resolve(options.token as string) : options.token;
 
+    if (!options.grpcClientOptions) options.grpcClientOptions = {};
+    options.grpcClientOptions["grpc.service_config"] = JSON.stringify({
+      methodConfig: [
+        {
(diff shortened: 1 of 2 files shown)


angelini:
> The problem is that receiving a message commits the RPC. This is discussed in [gRFC A6 Client Retries](https://github.com/grpc/proposal/blob/master/A6-client-retries.md#when-retries-are-valid). It mentions the Response-Headers, which are implicitly sent when the server responds with the first message.
Essentially, once gRPC has passed data back to the client there is no way to automatically retry. If gRPC retried how should it combine a new stream with what it has already responded with? Should it skip the first N responses? But what if the responses are different now? The problem is even worse for metadata (delivered via Response-Headers) as those cannot be provided to the client a second time.
gRPC is able to replay the client’s requests to multiple backends, but once it starts receiving a response from a backend it will become “fixed” to that backend and be unable to change its decision.
You will need application-level retry to re-establish the stream. When the client reestablishes the stream it may need to modify the request to inform the server what messages the client has already received.

https://stackoverflow.com/a/60939725

PR #11, opened 2022-10-06
Skip removed objects in queryBuilder

angelini:
In the case where we are building the query without considering removed objects, let's make the `possible_objects` CTE even simpler and not include them there as well.

This will benefit the API which often does `Get` request from version 0 so it does not consider removed objects.

### DIFF
--- a/internal/db/queryBuilder.go
+++ b/internal/db/queryBuilder.go
@@ -23,27 +23,38 @@ func newQueryBuilder(project int64, vrange VersionRange, objectQuery *pb.ObjectQ
 	}
 }
 
-func (qb *queryBuilder) possibleObjectsCTE() string {
-	return `
-		SELECT *
-		FROM dl.objects
-		WHERE project = __project__
-		  AND (
-		    -- live objects
-		    (
-		      start_version > __start_version__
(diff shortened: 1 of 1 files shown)


internal/db/queryBuilder.go
```diff
@@ -23,27 +23,38 @@ func newQueryBuilder(project int64, vrange VersionRange, objectQuery *pb.ObjectQ
...
-		    OR
-		    -- removed objects
-		    (
-		      start_version <= __stop_version__
-		      AND stop_version > __start_version__
-		      AND stop_version <= __stop_version__
-		    )
-		  )
-	`
+func (qb *queryBuilder) possibleObjectsCTE(includeRemovedObjects bool) string {
```
jasong689: nit: Could we name this `withRemovals` like we do with the method names? It was difficult to resolve that includeRemovedObjects == withRemovals for some reason 😆

jasong689 (approved):

internal/db/queryBuilder.go
angelini: Yeah that's much better :+1:

thegedge (approved):
🚀

PR #14, opened 2022-10-13
Always clone from source's version 0

jasong689:
We've decided that for cloning files over from one project to another, instead of specifying a version range we'll instead always clone from version 0 up to a specified version. This means any files in the target project that does not exist in source project should be removed. To do this we'll diff the existing files in the target with the source and set a stop version for them. To achieve this I needed to combine the queries from 2 different query builders (source and target) which meant updating how parameters are generated for each query.


angelini (approved):
🙌

PR #15, opened 2022-10-18
In-Memory Cache

angelini:
This is a second, complementary, caching system for DateiLager. A fun thing about content addressable storage: it's very easy to cache.

To begin let's cover what these two strategies are and how they work together, then we'll jump into the details about this PR.

1. Node local packed object caching:

This is the strategy added by @gmalette, where we cache packed objects that are referenced across many projects directly on our sandbox nodes. In Gadget's case, this is the most popular node modules.

Because these cached objects could be seen by multiple projects, we can only cache public items - namely `node_modules`. 

By caching them on the sandbox nodes directly, we can avoid reading these over the network. So when a project requests a common version of `typescript`, a packed object that is over 20M, we can send just a hash over the network and read it from a local disk instead of the network.

2. In-memory caching

This is what I've implemented in this PR. We create a read through cache on the DateiLager server nodes and use it to avoid reading common objects from Postgres.

We use an LRU cache that we can memory limit. Unlike the previous caching strategy, we can store public and private objects in this cache since it is only accessible behind DL's auth system.

In Gadget's case, we have lots of overlapping generated source files that would be nice to cache but since these objects are not necessarily packed they cannot be included in the node local cache.

----------------------------

Avoiding Postgres reads for common objects is an important feature of this PR, but it also allows us to better control our database queries and what resources we keep open during streaming read queries.

Today when you query DateiLager, the entire query is streamed. We open a DB transaction and begin to read objects from the database and writing them to the GRPC socket as soon as we can. This streaming process is quite important because it helps us limit memory consumption while reading potentially large objects from that database. During a request we only have a couple objects loaded into memory and we only go fetch more once the previous ones have been written to the stream.

One downside is that we keep our one big read query open with Postgres the entire duration of the DateiLager query. We stream results back only as quickly as the client can read them from the stream.

This PR breaks up querying Postgres into two steps:

1. We query all of the object hashes for our result set - and we can safely materialize this in memory since it doesn't contain any file contents.
 
2. And then in batches (currently of 50) we fetch content from the in-memory cache, or Postgres if we cache miss, and then write that batch to the stream before beginning a new batched read.
 
We split up our one big read from PG which included object metadata and their contents into first a metadata read and then chunked content reads.

----------------------------

In theory this feels like is should be a definite improvement, but I'm having a hard time crafting real world benchmarks that could be used to prove or disprove that theory.

I'm going to load this into Gadget for testing, but how else could we simulate how this may perform in production with multiple concurrent users.

I've got a load tests built with `ghz` (a GRPC load test tool) that can be run with:

```
$ make server

# And in a new shell
# 1. First setup the DB
$ make load-test-new && make load-test-update

# 2. Run the load test
$ make load-test-get
```

But I don't think my test is imitates the real world enough to show any improvements from this PR. 

/cc @airhorns

### DIFF
--- a/internal/db/content.go
+++ b/internal/db/content.go
@@ -3,13 +3,19 @@ package db
 import (
 	"bytes"
 	"context"
-	"encoding/hex"
 	"fmt"
 	"io"
 
+	"github.com/dgraph-io/ristretto"
 	"github.com/jackc/pgx/v5"
+	"github.com/jackc/puddle/v2"
 	"github.com/klauspost/compress/s2"
 	"github.com/minio/sha256-simd"
+)
+
+const (
+	MB              = 1000 * 1000
+	DecoderPoolSize = 1000
 )
 
 type Hash struct {
@@ -32,8 +38,27 @@ func (h *Hash) Bytes() []byte {
 	return hash
 }
 
+// Stolen from Go's standard library
+// And optimized for the Hash type
+const hextable = "0123456789abcdef"
+
 func (h *Hash) Hex() string {
-	return hex.EncodeToString(h.Bytes())
+	buffer := make([]byte, 64)
+	idx := 0
+
+	for _, v := range h.H1 {
+		buffer[idx] = hextable[v>>4]
+		buffer[idx+1] = hextable[v&0x0f]
+		idx += 2
+	}
+
+	for _, v := range h.H2 {
+		buffer[idx] = hextable[v>>4]
+		buffer[idx+1] = hextable[v&0x0f]
+		idx += 2
+	}
+
+	return string(buffer)
 }
 
 type ContentEncoder struct {
@@ -51,7 +76,7 @@ func NewContentEncoder() *ContentEncoder {
 	}
 }
 
-func (c *ContentEncoder) Encode(content []byte) ([]byte, error) {
+func (c *ContentEncoder) Encode(content DecodedContent) (EncodedContent, error) {
 	_, err := c.writer.Write(content)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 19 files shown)


internal/db/query.go
```diff
@@ -157,97 +113,157 @@ func filterObject(path string, objectQuery *pb.ObjectQuery, object *pb.Object) (
...
 
 	builder := newQueryBuilder(project, vrange, objectQuery)
-	sql, args := builder.build()
+	dbObjects, err := executeQuery(ctx, tx, builder)
+	if err != nil {
+		return nil, fmt.Errorf("get objects query, project %v vrange %v: %w", project, vrange, err)
+	}
 
-	rows, err := tx.Query(ctx, sql, args...)
-	closeFunc := func(_ context.Context) { rows.Close() }
```
thegedge, line 172: I guess this is no longer needed because we load up everything in some intermediate structure, and so it'll always be closed?

internal/db/project.go
```diff
@@ -39,6 +40,29 @@ func DeleteProject(ctx context.Context, tx pgx.Tx, project int64) error {
...
+		SELECT id, latest_version
+		FROM dl.projects
+	`)
+	if err != nil {
+		return nil, fmt.Errorf("snapshotProjects query: %w", err)
+	}
+
+	projects := []*pb.Project{}
+
+	for rows.Next() {
```
thegedge, line 54: Was gonna ask, but found the answer to my question about this and `Scan` returning an error below:

CleanShot 2022-10-31 at 14 13 36

internal/db/content.go
```diff
@@ -62,40 +87,146 @@ func (c *ContentEncoder) Encode(content []byte) ([]byte, error) {
...
+	decoders *puddle.Pool[*ContentDecoder]
+}
+
+func NewContentLookup() (*ContentLookup, error) {
+	cache, err := ristretto.NewCache(&ristretto.Config{
+		NumCounters: 100_000,
+		MaxCost:     1_000 * MB,
+		BufferItems: 64,
+	})
+	if err != nil {
```
thegedge, line 148: What would make this fail? Memory? Should we still boot the server if the cache can't be created, just without a cache?

internal/db/content.go
```diff
@@ -62,40 +87,146 @@ func (c *ContentEncoder) Encode(content []byte) ([]byte, error) {
...
+
+			if hashes[hash] {
+				decoded, err := decoder.Value().Decode(value)
+				if err != nil {
+					return nil, fmt.Errorf("cannot decode value from content table %v: %w", hash.Hex(), err)
+				}
+				contents[hash] = decoded
+			} else {
+				contents[hash] = value
+			}
```
thegedge: naming nit: I had to go to the caller to understand `hashes`, which made it hard to read this function on its own. Maybe just `shouldDecode`, or perhaps `unpacked` given what we passed in is `!dbObject.packed`?

internal/db/content.go
```diff
@@ -62,40 +87,146 @@ func (c *ContentEncoder) Encode(content []byte) ([]byte, error) {
...
+		for rows.Next() {
+			var hash Hash
+			var value []byte
+
+			err = rows.Scan(&hash.H1, &hash.H2, &value)
+			if err != nil {
+				return nil, fmt.Errorf("content lookup scan: %w", err)
+			}
+
+			cl.cache.Set(hash.Hex(), value, int64(len(value)))
```
thegedge, line 214: I almost commented on race conditions / write-through caching, but then I remembered the thing that's really important here is that everything is content addressable!

Just dropping the comment here in case you think we can make that clearer in any way.

thegedge (commented):
I think I got it all through my late afternoon 🧠 , but might be good to have you do a run through of the key bits tomorrow!

internal/db/query.go
angelini: Yeah, it feels a bit silly to have added this to remove it a couple days later. Oh well.

internal/db/project.go
angelini: Yeah this is why `rows.Close()` rarely leaves dangling query objects around, because we typically iterate over all of the results quite quickly.

That being said it is probably still good practice to add `defer rows.Close` after most queries in case we throw an error while scanning the rows.

internal/db/query.go
```diff
@@ -157,97 +113,157 @@ func filterObject(path string, objectQuery *pb.ObjectQuery, object *pb.Object) (
...
+	idx := 0
+	chunk_idx := 0
```
scott-rc: nit
```suggestion
	chunkIdx := 0
```

scott-rc (approved):
+1 to @thegedge nits and added one more.

Other than that - LGTM 👍

internal/db/content.go
angelini: The server will fail to start in this case, the error is checked before booting the server.

thegedge: I guess the main thing here is whether or not that would result in a crash loop. If it doesn't, we're good. If not, would that be worse than booting without the cache?

thegedge (approved):

internal/db/content.go
angelini: The server will fail to boot in that case, but we'll get a crash loop back off error and the deploy will fail - leaving the old version around

PR #16, opened 2022-10-18
Fixed array hash

angelini:
Change the `Hash` type from containing two slices (references to variable length arrays) to containing two fixed sized arrays.

This reduces a lot of pointer indirection by storing all of the data within the struct itself.

And, the reason I kicked off this PR, is that it allows you to directly compare two hashes `hash1 == hash2` instead of using `bytes.Equal` like we used to.

This built in equality check is useful because it will allow us to use `Hash` as a map key type. (https://golangbyexample.com/allowed-key-and-value-types-golang/)

However, this does mean we no longer support `nil` hashes, so our `queryBuilder` needs to return a 0 hash instead of `null` for cases when hashes don't make sense (for example with deleted objects).

### DIFF
--- a/internal/db/content.go
+++ b/internal/db/content.go
@@ -13,22 +13,22 @@ import (
 )
 
 type Hash struct {
-	H1 []byte
-	H2 []byte
+	H1 [16]byte
+	H2 [16]byte
 }
 
 func HashContent(data []byte) Hash {
 	sha := sha256.Sum256(data)
 	return Hash{
-		H1: sha[0:16],
-		H2: sha[16:32],
+		H1: *(*[16]byte)(sha[0:16]),
+		H2: *(*[16]byte)(sha[16:32]),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


internal/db/content.go
```diff
@@ -13,22 +13,22 @@ import (
...
+	H2 [16]byte
 }
 
 func HashContent(data []byte) Hash {
 	sha := sha256.Sum256(data)
 	return Hash{
-		H1: sha[0:16],
-		H2: sha[16:32],
+		H1: *(*[16]byte)(sha[0:16]),
+		H2: *(*[16]byte)(sha[16:32]),
```
angelini, line 24: This hack is only possible since Go 1.17 https://www.jetbrains.com/go/guide/tips/go-1-17-convert-slice-to-array-pointer/?playlist=new-features-in-goland-2021-2

scott-rc (approved):
LGTM 👍

PR #37, opened 2023-02-13
Add the ability to detect file patterns while rebuilding

angelini:
This adds file pattern matching to rebuild.

This will be used to detect if (and sometimes if and only if) certain files change, such as front end files.

We can do this without re-summarizing the files on disk by watching the files being untarred.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -6,12 +6,11 @@ linters:
   # https://golangci-lint.run/usage/linters/#enabled-by-default-linters
   enable:
     - depguard
-    - dogsled
+
     - errcheck
     - exportloopref
     - goconst
     - gocritic
-    - gocyclo
     - gofmt
     - goimports
     - goprintffuncname
@@ -25,13 +24,11 @@ linters:
     - typecheck
     - unconvert
     - unused
-    - varcheck
     - whitespace
-
-  # linters disabled because of go1.18 (see https://github.com/golangci/golangci-lint/issues/2649):
-  # - bodyclose
-  # - noctx
-  # - structcheck
+    - bodyclose
+    - noctx
+    - dogsled
+    - gocyclo
 
   # don't enable:
   # - wsl
@@ -41,5 +38,5 @@ linters:
 # Options for analysis running.
 run:
   skip-files:
-  - "internal/pb/*.pb.go$"
-  - internal/pb/object.go
+    - "internal/pb/*.pb.go$"
+    - internal/pb/object.go
--- a/internal/files/writer.go
+++ b/internal/files/writer.go
@@ -12,7 +12,29 @@ import (
 
 	"github.com/gadget-inc/dateilager/internal/db"
 	"github.com/gadget-inc/dateilager/internal/pb"
+	"github.com/gobwas/glob"
 )
+
+type FilePattern struct {
+	Iff     bool
+	pattern glob.Glob
+}
+
+func NewFilePattern(pattern string, iff bool) (*FilePattern, error) {
+	glob, err := glob.Compile(pattern)
+	if err != nil {
+		return nil, err
+	}
+
+	return &FilePattern{
+		Iff:     iff,
+		pattern: glob,
+	}, nil
+}
+
+func (f *FilePattern) Match(filename string) bool {
+	return f.pattern.Match(filename)
+}
 
 func fileExists(path string) bool {
 	_, err := os.Lstat(path)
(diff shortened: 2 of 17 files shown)


internal/files/writer.go
```diff
@@ -141,12 +166,20 @@ func WriteTar(finalDir string, cacheObjectsDir string, reader *db.TarReader, pac
...
+			return count, false, fmt.Errorf("read next TAR header: %w", err)
+		}
+
+		if pattern != nil && !patternMatch && pattern.Match(header.Name) {
+			patternMatch = true
+		}
+
+		if pattern != nil && pattern.Iff && patternExclusiveMatch && !pattern.Match(header.Name) {
+			patternExclusiveMatch = false
 		}
```
scott-rc: nit:
```suggestion
		if pattern != nil {
				if !patternMatch && pattern.Match(header.Name) {
					patternMatch = true
				}
		
				if pattern.Iff && patternExclusiveMatch && !pattern.Match(header.Name) {
					patternExclusiveMatch = false
				}
		}
```

internal/files/writer.go
```diff
@@ -12,8 +12,30 @@ import (
...
+		return nil, err
+	}
+
+	return &FilePattern{
+		Iff:     iff,
+		pattern: glob,
+	}, nil
+}
+
+func (f *FilePattern) Match(filename string) bool {
```
scott-rc, line 35: nit:
```suggestion
func (f *FilePattern) Matches(filename string) bool {
```

.golangci.yml
```diff
@@ -6,12 +6,11 @@ linters:
   # https://golangci-lint.run/usage/linters/#enabled-by-default-linters
   enable:
     - depguard
-    - dogsled
+
```
scott-rc, line 9: nit:
```suggestion
```

.golangci.yml
```diff
@@ -25,21 +24,19 @@ linters:
...
-  # - structcheck
+    - bodyclose
+    - noctx
 
   # don't enable:
   # - wsl
   # - gochecknoinits
   # - gomnd
+  # - dogsled
+  # - gocyclo
```
scott-rc: Why disable these?

pkg/client/client.go
```diff
@@ -336,36 +343,52 @@ func (c *Client) Rebuild(ctx context.Context, project int64, prefix string, toVe
...
 						return err
 					}
 
 					atomic.AddUint32(&diffCount, count)
+
+					if diffCount > 0 && pattern != nil {
+						// When we are doing an if and only if match, we want a falsy value to take precedence
+						// Otherwise, we want the truthy value to take precedence
+						if pattern.Iff && !didMatch {
+							patternMatch.CompareAndSwap(true, false)
```
scott-rc: What's the difference between `CompareAndSwap` and `Store` here?

pkg/cli/rebuild.go
```diff
@@ -65,6 +77,8 @@ func NewCmdRebuild() *cobra.Command {
 	cmd.Flags().StringVar(&dir, "dir", "", "Output directory")
 	cmd.Flags().StringVar(&ignores, "ignores", "", "Comma separated list of ignore paths")
 	cmd.Flags().BoolVar(&summarize, "summarize", true, "Should include the summary file (required for future updates)")
+	cmd.Flags().StringVar(&filePattern, "filepattern", "", "A glob file pattern which drives the patternDetected output boolean")
+	cmd.Flags().BoolVar(&filePatternIff, "filepatterniff", false, "Should the file pattern detection trigger if and only if those files have changed")
```
scott-rc: I hate this name, but I don't know an alternative so... 🤷

scott-rc (approved):
Bunch of small nits that you can ignore, but LGTM 👍

internal/files/writer.go
```diff
@@ -12,8 +12,30 @@ import (
 
 	"github.com/gadget-inc/dateilager/internal/db"
 	"github.com/gadget-inc/dateilager/internal/pb"
+	"github.com/gobwas/glob"
```
airhorns, line 15: > This library is created for compile-once patterns. This means, that compilation could take time, but strings matching is done faster, than in case when always parsing template.

👍  this is better than the other lib i was using, the compile every time thing bothered me about the one i was using

internal/files/writer.go
```diff
@@ -122,14 +144,17 @@ func makeSymlink(linkname string, path string) error {
 	return nil
 }
 
-func WriteTar(finalDir string, cacheObjectsDir string, reader *db.TarReader, packPath *string) (uint32, error) {
+func WriteTar(finalDir string, cacheObjectsDir string, reader *db.TarReader, packPath *string, pattern *FilePattern) (uint32, bool, error) {
```
airhorns, line 147: Does the glob library support chained globs like `foo/**,bar/**`? Or if we want multiple globs would we need to build it ourselves?

pkg/client/client.go
airhorns: God this must have been tricky to get right 🙏

airhorns (approved):
LGTM!

internal/files/writer.go
angelini: It supports `{foo,bar}*` maybe that's enough, if not we will need to build something ourselves.

angelini:
Cleaned up the implementation a good bit, but the functionality is the same.

PR #39, opened 2023-02-24
Remove webui

angelini:
This was an idea early on, but the recent move to extracting files to Bigquery is much more useful than what this provided.


scott-rc (approved):
🔥

PR #47, opened 2023-03-13
chore(gc): parameters are JSON serializable

MikePresman:
Similar to [this PR](https://github.com/gadget-inc/dateilager/pull/45) the change done here is related to the workflows in gadget. 

It turns out that not only do activity outputs need to be JSON serializable, but so do inputs. And as such this PR changes the parameters from bigint (not json serializable) to number.


angelini (approved):
Ahh that explains it

PR #57, opened 2023-05-02
Add worker count to span attributes

scott-rc:
This adds the number of workers we use to our span's attributes. This should make it easier to compare different counts in Honeycomb.


MikePresman (approved):

angelini (approved):
Makes sense :+1:

PR #64, opened 2023-05-27
Insert into the contents table outside the transaction

airhorns:
While working on parallel fixture loading gadget side, I encountered a number of dl project update failures caused by concurrent project updates deadlocking in DL's postgres. They look like this:

```
[2:58:36.496 PM] FATAL (dl-client) <cli/client.go:136>: command failed
    serverRole: "api"
    userVisible: false
    source: "platform"
    userspaceLogLevel: 20
    commandKey: "ServerOperation-oxYccZkKAEMY"
    trace_id: "27a3aa62d00cd3c96c4cd13788d52069"
    span_id: "1ae79b985ca4e495"
    trace_flags: "01"
    stacktrace: "github.com/gadget-inc/dateilager/pkg/cli.ClientExecute\n\tgithub.com/gadget-inc/dateilager/pkg/cli/client.go:136\nmain.main\n\tgithub.com/gadget-inc/dateilager/cmd/client/main.go:6\nruntime.main\n\truntime/proc.go:250"
    error: {
      "message": "update objects: close and receive fs.Update: rpc error: code = Internal desc = FS update packed objects for node_modules/string-similarity/: insert packed content, hash 5c910ba26299c8c3b9cfc5a4ca9b7734-9e3a52a5898872a8235f76467d6ef074: ERROR: deadlock detected (SQLSTATE 40P01)"
    }
```

I think that `ON CONFLICT DO NOTHING` helps resolve two writes happening at the same time for the same object, but if two transactions write different objects in opposite orders, they can still deadlock I believe.

Conveniently, I think that the contents table can actually be non-atomically committed to. Rows in it are immutable and garbage collected out of band already, so I think we can actually just add to the contents table outside the transaction. We ensure the contents are aded before the objects referencing them are, so we should never have objects referencing contents that don't exist. We might now have contents that aren't referenced by any objects as the contents committed but the objects didn't, but I think that's ok. It also means the performance sensitive objects transactions take fewer locks and ideally can commit quicker.

Vibing?

### DIFF
--- a/internal/testutil/db.go
+++ b/internal/testutil/db.go
@@ -6,11 +6,13 @@ import (
 
 	"github.com/gadget-inc/dateilager/internal/db"
 	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgconn"
 )
 
 type DbTestConnector struct {
-	conn *pgx.Conn
-	tx   pgx.Tx
+	conn    *pgx.Conn
+	tx      pgx.Tx
+	innerTx pgx.Tx
 }
 
 func newDbTestConnector(ctx context.Context, uri string) (*DbTestConnector, error) {
@@ -40,7 +42,16 @@ func (d *DbTestConnector) Connect(ctx context.Context) (pgx.Tx, db.CloseFunc, er
 	if err != nil {
 		return nil, nil, err
 	}
+	d.innerTx = innerTx
 	return innerTx, func(context.Context) {}, nil
+}
+
+func (d *DbTestConnector) Exec(ctx context.Context, sql string, args ...interface{}) (pgconn.CommandTag, error) {
+	if d.innerTx != nil {
+		return d.innerTx.Exec(ctx, sql, args...)
+	} else {
+		return d.tx.Exec(ctx, sql, args...)
+	}
 }
 
 func (d *DbTestConnector) close(ctx context.Context) {
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -4,10 +4,12 @@ import (
 	"context"
 
 	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgconn"
 )
 
 type CloseFunc func(context.Context)
 
 type DbConnector interface {
+	Exec(ctx context.Context, sql string, args ...interface{}) (pgconn.CommandTag, error)
 	Connect(context.Context) (pgx.Tx, CloseFunc, error)
 }
--- a/internal/db/update.go
+++ b/internal/db/update.go
@@ -39,7 +39,7 @@ func DeleteObject(ctx context.Context, tx pgx.Tx, project int64, version int64, 
 }
 
 // UpdateObject returns true if content changed, false otherwise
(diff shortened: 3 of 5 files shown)


scott-rc (approved):
Gonna let Alex make the final call, but makes sense and LGTM 👍

angelini:
This makes sense to me, but has never been something we've run into in testing or production.

I guess because we typically only have 1 DL writer to an environment at once and if we do parallelize writes we never hit this case? 🤷‍♂️ 

I have no problem with this change, I'm just going to try and reproduce it locally.

airhorns:
This is the PR which produced it Gadget side for me https://github.com/gadget-inc/gadget/pull/6835 if you flip that back to the most recent release of DL it might take a few tries but resetting fixtures will cause it

airhorns:
@angelini what would you like to do with this one? Can you see a downside such that it feels risky?

internal/testutil/db.go
```diff
@@ -40,9 +42,18 @@ func (d *DbTestConnector) Connect(ctx context.Context) (pgx.Tx, db.CloseFunc, er
 	if err != nil {
 		return nil, nil, err
 	}
+	d.innerTx = innerTx
 	return innerTx, func(context.Context) {}, nil
 }
 
+func (d *DbTestConnector) Exec(ctx context.Context, sql string, args ...interface{}) (pgconn.CommandTag, error) {
+	if d.innerTx != nil {
+		return d.innerTx.Exec(ctx, sql, args...)
```
angelini, line 51: Doesn't this mean `db.Exec()` is going to use the `innerTx` inside any test that is testing `UpdateObject` since we'll have called `db.Connect()`.

Would we not want to always use `d.tx.Exec()` for all calles to `DbTestConnector.Exec()`?

angelini:
The only issue that comes to mind is the fact that `UpdateObject()` now sends two queries per call instead of one batch with 3 queries in it.

This method is called once per packed object and once per object while updating a project (which can be quite a large number). I've been meaning to turn this into a fully batched call so that we can update multiple objects at once but until then this is the bottleneck of updating a project.

angelini (approved):

internal/testutil/db.go
airhorns: I kinda assumed that the reason the test connector was separate was so that all tests were wrapped in a transaction that we could revert at the end right? So it has to use the transaction, and not just any old connection? I tried with `d.tx.Exec()` but that panic'd, I think because pgx doesn't model nested transactions as a tree of children, but instead a stack, and if you try to use anything other than the top item in the stack, its an error. Thats how postgres actually works under the hood as well -- it's just savepoints within the same transaction as opposed to independent chunks that can evolve separately. Is that going to create test fuckery though? I dont know the codebase well enough to say.

angelini: > I kinda assumed that the reason the test connector was separate was so that all tests were wrapped in a transaction that we could revert at the end right?

Yeah that's right. Since we rollback `d.tx` after each test I thought it would be fine.

> I tried with d.tx.Exec() but that panic'd, I think because pgx doesn't model nested transactions as a tree of children, but instead a stack, and if you try to use anything other than the top item in the stack

But yeah I didn't realize it wasn't actually a tree structure. Just played around with this and you're right. :+1:

PR #65, opened 2023-08-03
Add project rollback

angelini:
This is needed by the AAC project to rollback in case of problems with the AAC migration.


js/spec/grpc-client.spec.ts
```diff
@@ -52,6 +52,38 @@ describe("grpc client operations", () => {
     expect(receivedObjects).toEqual(objects);
   });
 
+  it("can rollback a project", async () => {
```
danroberts, line 55: One situation that I'm not sure about what would happen is: 

- The user rollsback to v1, then makes a new change. Are we at version == 3 now? does version == 2 still exist? if so, how does that play into the diffing logic that dl does when we ask it to roll forward to a new version? 

My mental model is a bit hazy, but I kind of thought that it stored diffs, so that we might catch the v1 -> v2 diff if we ask to go from v1 -> v3? 

Could we add a test to clarify what is expected here?

danroberts: Maybe this case is covered the the go test below, but my go is not good enough to really understand it.

danroberts (approved):
Hard for me to review the go code, but thank you for fitting this in!

scott-rc (approved):
LGTM 👍

js/spec/grpc-client.spec.ts
angelini: If you rollback from 3 to 1, and then add a file, you'll be back to 2.

This works because the `Reset()` function we're using internally throws away all diffs between after version 1.

So it is safe to re-use version 2.

PR #67, opened 2023-08-09
File match include exclude

angelini:
In order to fix HMR we need to properly detect changes in `frontend/**` but not `frontend/vite.config.js`.

Annoyingly, there's no way to express that in either the glob library we were using or in a Go Regex (they do not support negative lookaheads).

Instead this changes the syntax asking for an include matcher and/or an exclude matcher.

The resulting `fileMatch` is true if and only if all updated files match the `include` matcher and none of them match the `exclude` matcher.

I also added an explicit Gadget file matcher integration test, @danroberts let me know if I need to add more examples to that.

### DIFF
--- a/internal/files/writer.go
+++ b/internal/files/writer.go
@@ -16,25 +16,47 @@ import (
 	"github.com/gobwas/glob"
 )
 
-type FilePattern struct {
-	Iff     bool
-	pattern glob.Glob
+type FileMatcher struct {
+	include *glob.Glob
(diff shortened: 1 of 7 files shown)


angelini:
This is slightly breaking, but as long as we change `filePattern` to `fileMatch` in Gadget, it should be an easy deployment.

scott-rc (approved):
LGTM 👍

PR #72, opened 2023-12-12
change: add a more clear error for duplicate project creation

MikePresman:
Whenever a project is attempted to made again the database throws a uniqueness error. This usually comes not so clearly out clientside and as such this change implements a cleaner error message for handling recreating projects that may occur in instances such as job retries with multiple steps.

### DIFF
--- a/internal/db/project.go
+++ b/internal/db/project.go
@@ -2,6 +2,7 @@ package db
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/gadget-inc/dateilager/internal/pb"
@@ -13,6 +14,13 @@ func CreateProject(ctx context.Context, tx pgx.Tx, project int64, packPatterns [
 		INSERT INTO dl.projects (id, latest_version, pack_patterns)
 		VALUES ($1, 0, $2)
 	`, project, packPatterns)
+
+	var projectExistsError = errors.New("ERROR: duplicate key value violates unique constraint \"projects_pkey\" (SQLSTATE 23505)")
+
+	if err != nil && err.Error() == projectExistsError.Error() {
+		return fmt.Errorf("project id already exists")
+	}
+
 	if err != nil {
 		return fmt.Errorf("create project %v, packPatterns %v: %w", project, packPatterns, err)
 	}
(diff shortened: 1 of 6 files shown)


internal/db/project.go
```diff
@@ -13,6 +14,13 @@ func CreateProject(ctx context.Context, tx pgx.Tx, project int64, packPatterns [
 		INSERT INTO dl.projects (id, latest_version, pack_patterns)
 		VALUES ($1, 0, $2)
 	`, project, packPatterns)
+
+	var projectExistsError = errors.New("ERROR: duplicate key value violates unique constraint \"projects_pkey\" (SQLSTATE 23505)")
+
+	if err != nil && err.Error() == projectExistsError.Error() {
+		return fmt.Errorf("project id already exists")
+	}
```
scott-rc, line 22: Should we just return `nil` and noop instead?

angelini: Well that can be dangerous id the `packPatterns` is different, in that case you could think you are making a project with one set of values and instead another was already created.

angelini:
Can you also update the JS `grpc-client` to throw a nice error like `ProjectAlreadyExists`

internal/db/project.go
scott-rc: right, makes sense 👍

js/src/grpc-client.ts
```diff
@@ -119,17 +119,25 @@ export class DateiLagerGrpcClient {
...
-      () => this._client.newProject({ id: project, packPatterns, template }, this._rpcOptions())
-    );
+        () => this._client.newProject({ id: project, packPatterns, template }, this._rpcOptions())
+      );
+    } catch (error) {
+      if (error instanceof RpcError && error.code == "ALREADY_EXISTS") {
+        throw new ProjectAlreadyExistsError(`project id ${project} already exists`);
+      } else {
+        throw error;
+      }
```
scott-rc: nit: no need for an `else` block here

```suggestion
      if (error instanceof RpcError && error.code == "ALREADY_EXISTS") {
        throw new ProjectAlreadyExistsError(`project id ${project} already exists`);
      }
      throw error;
```

js/src/utils/errors.ts
```diff
@@ -0,0 +1,5 @@
+export class ProjectAlreadyExistsError extends Error {
+  constructor(msg: string) {
+    super(msg);
+  }
```
scott-rc: nit: no need to re-implement the same constructor

```suggestion
```

scott-rc (approved):
LGTM 👍

angelini (approved):

PR #77, opened 2024-03-09
Don't terminate DL binary client subprocesses when the parent process is SIGTERM'd

airhorns:
We were seeing failures in Gadget during deploys where a running pod would be doing DL work and get SIGTERM'd by kubernetes. Normally, unixes will only deliver that signal to the parent process, but `execa` takes care to propagate SIGTERMs from the parent to any spawned children. This has the effect of prematurely killing a dl binary client subprocess mid operation. We use graceful shutdown in gadget to allow currently-running operations within our workers to finish, and so to allow that, we don't want these signals propagated.

### DIFF
--- a/js/src/binary-client.ts
+++ b/js/src/binary-client.ts
@@ -374,6 +374,7 @@ export class DateiLagerBinaryClient {
     }
     const subprocess = execa(this._options.command, baseArgs.concat(args), {
       cwd,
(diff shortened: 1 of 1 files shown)


angelini (approved):

angelini:
@airhorns I'm gonna manage deploying a new DL version and getting this into Gadget.

PR #86, opened 2024-05-04
Build dateilager within the dockerfile using the nix environment

airhorns:
This switches the build process to:
 - build DL within the dockerfile for repeatable builds, instead of copying in from outside
 - use the nix environment to setup the env for the build, so it's the same go version / protoc version etc etc in and outside of docker
 - use a multistage docker build so we have slim containers with just the binary at the end

I did all this because I was fighting to get a docker image built for linux_arm64, which is what orbstack needs to run the image locally in a k8s setup! It was annoying. This also adds linux_arm64 as a build target to the makefile, and teaches the dockerfile how to build the binary for the target architecture that the dockerfile is building for.

Yeesh.


airhorns:
* **#91** <a href="https://app.graphite.dev/github/pr/gadget-inc/dateilager/91?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#89** <a href="https://app.graphite.dev/github/pr/gadget-inc/dateilager/89?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#85** <a href="https://app.graphite.dev/github/pr/gadget-inc/dateilager/85?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#90** <a href="https://app.graphite.dev/github/pr/gadget-inc/dateilager/90?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#86** <a href="https://app.graphite.dev/github/pr/gadget-inc/dateilager/86?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#87** <a href="https://app.graphite.dev/github/pr/gadget-inc/dateilager/87?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @airhorns and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

Makefile
```diff
@@ -21,24 +21,14 @@ PROTO_FILES := $(shell find internal/pb/ -type f -name '*.proto')
...
 
-install:
-	go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.28.1
-	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.2
-	go install github.com/grpc-ecosystem/grpc-health-probe@v0.4
-	go install -tags 'postgres' github.com/golang-migrate/migrate/v4/cmd/migrate@v4.15
-	go install github.com/bojand/ghz/cmd/ghz@v0.110.0
-	go install github.com/gadget-inc/fsdiff/cmd/fsdiff@v0.4
-	go install github.com/stamblerre/gocode@latest
-	go install golang.org/x/tools/cmd/goimports@latest
```
airhorns, line 40: Stuck these in the nix env, and then removed the ones that now come bundled with vscode (gocode => gopls, goimports is bundled also). This way, the versions used are the same in dev and in docker

Dockerfile
```diff
@@ -20,8 +33,11 @@ RUN mkdir -p /home/main/secrets
 VOLUME /home/main/secrets/tls
 VOLUME /home/main/secrets/paseto
 
-COPY release/server_linux_amd64 server
+COPY --from=build-stage /app/release/server_linux_${TARGETARCH} server
```
airhorns, line 51: Did the nice multistage thing so all the build deps aren't in the final image

flake.nix
```diff
@@ -63,6 +64,27 @@
...
+              flake.packages.postgresql
+              flake.packages.dev
+              flake.packages.clean
+              git
+              protobuf
+              protoc-gen-go
+              protoc-gen-go-grpc
+              go-migrate
+              mkcert
+            ];
```
airhorns, line 81: This makes the devShell build all the stuff you need to work on DL into your shell, but doesn't build DL itself. I always found that annoying with the original setup that I think Kira created that cd-ing into the dir built the whole project, which doesn't work so good when you are working on it and it might be busted.

Dockerfile
```diff
@@ -1,17 +1,34 @@
-FROM --platform=linux/amd64 registry.fedoraproject.org/fedora-minimal:40
+FROM nixos/nix:2.22.0 AS build-stage
```
angelini, line 1: Without passing `--platform` here (or when calling docker build) I'm pretty sure this will build a darwin executable. (Unless orbstack handles that differently).

This caused us issues in the past where we ended up deploying Darwin executables to prod.

Dockerfile
```diff
@@ -1,17 +1,34 @@
...
+filter-syscalls = false
+EOF
+
+WORKDIR /app
+
+COPY flake.nix flake.lock  ./
+COPY development ./development
+
+# build the nix environment once as an early docker layer
+RUN nix develop -c true
```
angelini: There's nothing we can call to avoid these `nix develop -c` prefixes is there?

Dockerfile
```diff
@@ -1,17 +1,34 @@
...
+
+# copy the go modules and download em
+COPY go.mod go.sum ./
+RUN nix develop -c go mod download
+
+# copy everything else and build the project
+COPY . ./
+RUN nix develop -c make release/server_linux_$TARGETARCH
+
+FROM buildpack-deps:bullseye AS build-release-stage
```
angelini, line 32: This drops a lot of the debugging tools that have saved our butts in prod before:

```
RUN microdnf install -y curl findutils gzip iputils less postgresql procps shadow-utils tar time which \
    && microdnf clean all
```

Some of these might be built-in `bullseye` but it would suck if they weren't there when we ran into problems.

flake.nix
angelini: I'm down to properly Nix convert the project, it was in a very half converted state for a while.

But I do really miss the ability to control package versions, if ever Nix updates a package before we want to we're SOL.

Makefile
```diff
@@ -206,8 +201,10 @@ upload-prerelease-container-image: release
 	docker build -t gcr.io/gadget-core-production/dateilager:$(GIT_COMMIT) .
 	docker push gcr.io/gadget-core-production/dateilager:$(GIT_COMMIT)
```
angelini: I suspect that if we called this upload function (which should ultimately live in a Github Actions) it will push an ARM image to our registry.

.dockerignore
```diff
@@ -0,0 +1,9 @@
+.git
+.direnv
+Dockerfile
+js
+test
+tmp
+bin
+input
+node_modules
```
angelini: `node_modules` should be within the `js` dir so we shouldn't need to repeat it.

I'd also probably ignore the `rebuild` directory since the docs use that to test rebuilding into (or we could change the docs to point to `tmp/rebuild`)

We also likely want to ignore `bin` and `release` which gets created when binaries are built.

And we can probably ignore `docs` and `.vscode` as well.

angelini: Should we maybe force the certs and key files from `development` to be ignored so we never accidentally deploy the service using dev keys.

flake.nix
airhorns: It's possible in nix, it's just annoying! You add a new dep at the top of a file for a nixpkgs sha that has the version of the thing you actually want, and then you refer to that version of nixpkgs in your exported packages! We did it in gadget a couple times when we didn't want to upgrade things.

Dockerfile
airhorns: I was thinking that we'd want to use the evil-ssh-pod for that, but I am not super experienced with the DL deployment so if you find them useful in the pod I'm totally fine with that, it just means a bigger container but whatever

airhorns: Updated to leave these in

Makefile
airhorns: You're right! The whole reason I was messing around with this is to be able to build a local image for testing locally, but also push a prerelease image to the real registry for use in Gadget's CI to use as an e2e test of the whole thing. Locally we need arm but CI/prod we need amd64. I had multiarch builds set up in the downstream branch but didn't git-fu it up here in this one, moved it now so its its final form on this branch

.dockerignore
airhorns: Word, ignored all of `development` except the nix stuff

Dockerfile
airhorns: See comment below -- it's set up for multiarch now in the makefile

PR #93, opened 2024-06-03
db(add columns): created_at

MikePresman:
Added these columns since they are extremely helpful for tracking when a version was created


migrations/000007_add_date_columns_to_objects.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE dl.objects ADD COLUMN created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL;
```
angelini, line 1: I think defaulting to current timestamp is cool, we wouldn't need to add any code changes and would get created_at for free.

The only worry here is what do we do for existing rows, I kind of feel like they should be null instead of some fixed date when we rolled this out.

Can you try testing the following:

- Create rows without this patch
- Run this migration (without the `NOT NULL` part)

Do the existing rows get the date of the migration as their value or null? I suspect they'll get the date of the migration.

We might need to run a first migration with default null and then a second migration with `default current_timestamp`

PR #100, opened 2025-01-08
faster hardlink with fastwalk

udnay:
Use the [fastwalk](https://github.com/charlievieth/fastwalk) library that walks a directory in parallel.

The `os.Link` that creates a hard link between two files can only do one file in a directory at a time, so using something like `io_uring` won't help because the parent directory acquires a lock and forces a concurrency of one. `fastwalk` will walk directories in parallel so links created in distinct directories will be able to occur in parallel.

Benchmark numbers run on `tempfs`

Implementation | ns/op
-- | --
Current | 31,465,946
Fastwalk | 16,020,991
Iouring-fastwalk | 64,682,665

### DIFF
--- a/internal/files/writer.go
+++ b/internal/files/writer.go
@@ -214,7 +214,6 @@ func HardlinkDir(olddir, newdir string) error {
 
 	fastwalkConf := fastwalk.DefaultConfig.Copy()
 	fastwalkConf.Sort = fastwalk.SortDirsFirst
-	dirPermissions := rootInfo.Mode()
 
 	return fastwalk.Walk(fastwalkConf, olddir, func(oldpath string, d os.DirEntry, err error) error {
 		if err != nil {
@@ -223,8 +222,18 @@ func HardlinkDir(olddir, newdir string) error {
 
 		newpath := filepath.Join(newdir, strings.TrimPrefix(oldpath, olddir))
 
+		// The new "root" already exists so don't recreate it
+		if newpath == newdir {
+			return nil
+		}
+
 		if d.IsDir() {
-			err = os.Mkdir(newpath, dirPermissions)
+			info, err := d.Info()
+			if err != nil {
+				return fmt.Errorf("unable to get directory info %v: %w", oldpath, err)
+			}
+
+			err = os.Mkdir(newpath, info.Mode())
 			if err != nil {
 				return fmt.Errorf("cannot create dir %v: %w", newpath, err)
 			}
--- a/internal/files/writer.go
+++ b/internal/files/writer.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/charlievieth/fastwalk"
 	"github.com/gadget-inc/dateilager/internal/db"
 	"github.com/gadget-inc/dateilager/internal/pb"
 	"github.com/gobwas/glob"
@@ -211,7 +212,10 @@ func HardlinkDir(olddir, newdir string) error {
 		return fmt.Errorf("cannot create new root dir %v: %w", olddir, err)
 	}
 
-	return filepath.WalkDir(olddir, func(oldpath string, d os.DirEntry, err error) error {
+	fastwalkConf := fastwalk.DefaultConfig.Copy()
+	fastwalkConf.Sort = fastwalk.SortDirsFirst
+
+	return fastwalk.Walk(fastwalkConf, olddir, func(oldpath string, d os.DirEntry, err error) error {
 		if err != nil {
 			return fmt.Errorf("failed to walk dir: %v, %w", oldpath, err)
 		}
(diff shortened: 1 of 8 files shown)


internal/files/writer.go
```diff
@@ -201,15 +202,34 @@ func HardlinkDir(olddir, newdir string) error {
...
-		if info.IsDir() {
-			err := os.MkdirAll(newpath, info.Mode())
+		if d.IsDir() {
+			info, err := d.Info()
+
+			if err != nil {
+				return fmt.Errorf("unable to get directory info %v: %w", oldpath, err)
+			}
+
+			err = os.MkdirAll(newpath, info.Mode())
```
airhorns: This is like `mkdir -p` right, which is probs slower than just a straight mkdir cause it has to stat all the parents every time? I think it was likely being defensive against the parent dirs not existing yet, but knowing what we know now, I think it would be safe to create the root newdir once before starting the walk, and then making this just do a straight mkdir?

Also, does `info.Mode()` do a separate stat syscall to get its info? Any way to avoid that if so?

airhorns (approved):
LGTM after comments! We can also build confidence by cutting a pre-release and running it through the gadget CI

internal/files/writer.go
udnay: Yeah, we sort with list dir's first so we know that we create the parents first afaik, I thought we only called `mkdirall` on the root. I'll fix it up

udnay: For `info.Mode()` it looks like it doesn't do extra syscalls looking at the implementations. It's a 64bit int that is masked to indicate the permissions. We could just copy and use the the `rootDir`'s info and pass that into the walkDir. I assume that the child directories can have the same permissions as the parent.

I don't know enough about the file structure, would there be a case where a child directory would be less permissive than the parent or more permissive? Seems like it would be pretty safe to just use the root's permissions for the directory and save on the costs there. Using the root would save us on the `lstat` call in `dirEntry.info()`

internal/files/writer.go
```diff
@@ -201,15 +202,29 @@ func HardlinkDir(olddir, newdir string) error {
...
+	}
+
+	err = os.MkdirAll(newdir, rootInfo.Mode())
+	if err != nil {
+		return fmt.Errorf("cannot create new root dir %v: %w", olddir, err)
+	}
+
+	fastwalkConf := fastwalk.DefaultConfig.Copy()
+	fastwalkConf.Sort = fastwalk.SortDirsFirst
+	dirPermissions := rootInfo.Mode()
```
udnay: We take the rootInfo and pass it into the walk func

internal/files/writer.go
```diff
@@ -201,15 +202,34 @@ func HardlinkDir(olddir, newdir string) error {
...
-			return fmt.Errorf("failed to walk dir: %v, %w", info, err)
+			return fmt.Errorf("failed to walk dir: %v, %w", oldpath, err)
 		}
 
 		newpath := filepath.Join(newdir, strings.TrimPrefix(oldpath, olddir))
 
-		if info.IsDir() {
-			err := os.MkdirAll(newpath, info.Mode())
+		// The new "root" already exists so don't recreate it
+		if newpath == newdir {
```
udnay, line 226: Turns out `MkdirAll` has a handy check to make sure that a directory doesn't already exist which we hit when we start the walk. Since we create the root above we can just quickly return now that we do `Mkdir`

internal/files/writer.go
airhorns: I think that is maybe fine, but the risk of it not being fine is not worth it IMO. I am not super familiar with how picky `npm` and/or `yarn` are about the permissions on the various files and folders they manage, but I do know at least some of the bits need their executable bit set, and so I'd vote for maintaining the interface we had before. if we still see this as a big bottleneck later i think we could do some proving that the perms are fine to just set to consistent values, or we could have the DL server ship us the exact perms of all the stuff alongside the tarball so we can just have it in memory. 

And ok 👍  with the other bits!

internal/files/writer.go
```diff
@@ -201,15 +202,34 @@ func HardlinkDir(olddir, newdir string) error {
...
+			return nil
+		}
+
+		if d.IsDir() {
+			err = os.MkdirAll(newpath, dirPermissions)
```
airhorns: Isn't this the one we want to be a plain `mkdir` since its within the walk and will be oft repeated?

udnay: Derp, copy paste mistake. Will fix.

internal/files/writer.go
udnay: The file permissions should be the same since we're hard linking them, this is only the dir creation. Likely dir creation is much less frequent than files. But better safe than sorry.

airhorns:
Send it!

PR #104, opened 2025-02-05
Use overlay for cached

udnay:
Use an overlay for dateilager's cached service. Instead of giving each sandbox pod its own hard linked copy of the `dl_cache` we will mount an overlayfs of the `cached`'s `dl_cache`

In the pod that's mounting to overlay we can see that we can hardlink between the 2 directories and that the pod can write to the cache and corrupt it.
Screenshot 2025-02-05 at 4 27 01 PM

however on the `cached` pod the contents are still fine
 
Screenshot 2025-02-05 at 4 27 18 PM

and additionally if we mount a second pod, it too sees the clean copy.
Screenshot 2025-02-05 at 4 29 40 PM

The pods are configured to use their ephemeral space so any writes to `dl_cache` will be cleaned up when the pod is killed.

The `cached` pod stores the "golden cache" at `/var/lib/kubelet/dateilager_cache`

For a simple example pod lets look at
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: busybox-csi
spec:
  containers:
  - name: busybox
    image: busybox
    command: 
      - "/bin/sh"
      - "-c"
      - "sleep infinity"
    volumeMounts:
    - name: gadget
      mountPath: /gadget
      subPath: gadget
  volumes:
  - name: gadget
    csi:
      driver: com.gadget.dateilager.cached
      volumeAttributes:
        placeCacheAtPath: "dl_cache"
        mountCache: "true"
      readOnly: false
```
A pod that is mounting a volume managed by the dateilager-csi will get a new ephemeral dir (like `emptyDir`) that is on the host `/var/lib/kubelet/pods/88064d87-6870-4da5-8ba1-d9ac62b3f3ff/volumes/kubernetes.io~csi/gadget/mount` this is the `targetPath` that comes in on the `csi.NodePublishVolumeRequest`

The directory structure that we create inside of the target dir is like this
```
/var/lib/kubelet/pods/
   | -- .../
      | -- gadget/mount/
          | -- gadget/
          | -- gadget_writeable/
          | -- work/
```
`gadget` is the target mount point for the overlay. `gadget_writeable` is the `upperdir` and `work` is the `workdir`

The readonly `lowerdir` is `/var/lib/kubelet/dateilager_cache` which contains `dl_cache`.


pkg/api/cached.go
```diff
@@ -167,11 +174,74 @@ func (c *Cached) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
...
+		}
 
+		mountArgs := []string{
+			"-t",
+			"overlay",
+			"overlay",
+			gadgetDir,
+			"-n",
+			"-o",
+			fmt.Sprintf("lowerdir=%s,upperdir=%s,workdir=%s", c.StagingPath, upperdir, workdir),
```
airhorns: We should consider what value we want for the `redirect_dir` option. When I was working with overlayfs for making the gadget CI bits that reset the filesystem faster, I had to set it to `on`, but that's not necessarily true here. https://docs.kernel.org/filesystems/overlayfs.html#redirect-dir

See https://github.com/gadget-inc/gadget/pull/11070/files#diff-37490d715785dc29d4323daef03e84518b414f5fdca4d258c9f415c936080d4f if you are curious

udnay: TIL. 
We can set it, I'm not sure how often we'll change the cache but it'll be nice to have the option.

What do you mean by 
> CI bits that reset the filesystem faster

As in, `overlay` was faster or something else?

udnay:
@airhorns https://github.com/gadget-inc/gadget/pull/14578 CI is passing with the overlay running. I think this PR is ready for review.

pkg/api/cached.go
```diff
@@ -145,8 +151,9 @@ func (c *Cached) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
 
 	var cachePath string
 	var targetPermissions os.FileMode
-
-	if suffix, exists := volumeAttributes["placeCacheAtPath"]; exists {
```
airhorns, line 149: Should make this attribute mandatory since we're always gonna mount at a subpath

pkg/api/cached.go
```diff
@@ -167,11 +174,74 @@ func (c *Cached) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
...
 	}
 
-	version, err := c.writeCache(cachePath)
-	if err != nil {
-		return nil, err
-	}
+	// Perform an overlay mount if the mountCache attribute is true
+	// Mount the root `StagingPath` at `targetPath/gadget` making `targetPath/gadget` `0777`
+	if mountCache, exists := volumeAttributes["mountCache"]; exists && mountCache == "true" {
+		upperdir := path.Join(targetPath, "gadget_writeable")
```
airhorns: We gonna try to avoid this extra subfolder if we can after talking IRL

.github/workflows/test-integration.yml
```diff
@@ -0,0 +1,31 @@
+name: Test Integration
```
udnay, line 1: Added this because overlay won't run on mac's so bumping the csi-tests out to be more integration sanity tests.

pkg/api/cached.go
```diff
@@ -287,3 +355,13 @@ func getFolderSize(path string) (int64, error) {
 	})
 	return totalSize, err
 }
+
+func execCommand(cmdName string, args ...string) error {
```
udnay, line 358: In github actions we're not running as root, in our deployment we run as root in a privileged container so this is how we can call `mount`/`umount` during our tests. If there's a more elegant way to do this let me know.

pkg/api/cached.go
```diff
@@ -140,39 +156,81 @@ func (c *Cached) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
...
+		err = os.Chmod(upperdir, 0777)
+		if err != nil {
+			return nil, fmt.Errorf("failed to change permissions of overlay upper directory %s: %v", upperdir, err)
+		}
 	}
 
-	if err := os.MkdirAll(targetPath, targetPermissions); err != nil {
-		return nil, fmt.Errorf("failed to create target directory %s: %s", targetPath, err)
+	workdir := path.Join(volumePath, WORK_DIR)
+	err = os.MkdirAll(workdir, 0777)
```
udnay, line 181: This did not seem to set the dir to `0777` so I had to add the explicit `chmod` below.

udnay:
@airhorns I made the changes, I still have some clean up to do but I think this is ready for a quick second pass.

pkg/api/cached.go
```diff
@@ -58,17 +62,29 @@ func (c *Cached) PopulateDiskCache(ctx context.Context, req *pb.PopulateDiskCach
...
+	// Once we've prepared the cache make it read-only for
+	// everyone except the user running the daemon
+	err = os.Chmod(c.GetCachePath(), 0755)
+	if err != nil {
+		return fmt.Errorf("failed to change permissions of cache path %s: %v", c.GetCachePath(), err)
+	}
+
 	c.currentVersion = version
 
+	logger.Info(ctx, c.GetCachePath())
```
airhorns: want to 🔥 or make nice

pkg/api/cached.go
airhorns: Terrifying -- maybe because its a mkdirall, the new perms only apply if it is creating new dirs, but it leaves existing ones alone?

pkg/api/cached.go
airhorns: I dont know of one 🤷

pkg/api/cached.go
```diff
@@ -140,39 +156,81 @@ func (c *Cached) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
...
+		return nil, fmt.Errorf("failed to create target path directory %s: %v", targetPath, err)
+	}
+
+	mountArgs := []string{
+		"-t",
+		"overlay",
+		"overlay",
+		"-n",
+		"--options",
+		fmt.Sprintf("redirect_dir=on,lowerdir=%s,upperdir=%s,workdir=%s", c.StagingPath, upperdir, workdir),
```
airhorns, line 209: Did redirect_dir end up being necessary?

airhorns (approved):
LGTM, or definitely good enough to canary

scott-rc (approved):

pkg/api/cached.go
udnay: I thought I caught all of my debug statements.

pkg/api/cached.go
udnay: Not entirely, it seems to be useful if there are changes to the lower dir that files don't disappear which seems good in general.

PR #105, opened 2025-02-11
Yandu/use overlay only

udnay:
This PR enables the `cached` client to mount and use an overlayfs for the dl cache. The main benefit is that we will not need to copy all of the cache's files into each pod at startup. The amount of data written will be `O(num dependencies)` vs `O(all files in the cache)`

### What happens
`cached` has a local copy of the dependency cache on the node that it is running on. When a new node requires the cache dependencies `cached` will use the existing cache as the `lowerDir` for an overlay mount. The mount location is managed under the `/var/lib/kubelet/pods/<uuid>/storage/.../` ephemeral location that is similar to where an `emptyDir` would be placed, this location is the `targetPath` inside of the `NodePublishVolume` function.

`cached` will create an `upperDir` and `workDir` in `targetPath/..` which is an available workspace that is not presented to the pod, only `targetPath` will be mounted/visible to the pod.

So the node running will be presented with the same data structure as it expected before with a `dl` and `app` directory in their pod. (**It might be worth it to make this less gadget specific**).

After the mounting everything works as before, dl will hardlink any dependencies into the `app`'s `node_module` directory. The one drawback with this current approach is that a hardlink is effectively a `write` in overlayfs, so while we do not require the full cache to be present in all pods, any needed file will be copied "up" to the writeable `upperDir` when linked.

### Other pieces to the PR
- I've added some k8s helpers so that `cached` can be run against a local k8s cluster
- Integration tests that will only run in CI/linux family VM as overlay is not supported on a Mac


result
```diff
@@ -0,0 +1 @@
+/nix/store/zf81z4pfl7nb064zjignm6jy94q8b121-go-1.22.12
```
scott-rc: What's this for?

scott-rc (approved):
LGTM 👍

pkg/api/cached.go
```diff
@@ -58,17 +62,29 @@ func (c *Cached) PopulateDiskCache(ctx context.Context, req *pb.PopulateDiskCach
...
+	// Once we've prepared the cache make it read-only for
+	// everyone except the user running the daemon
+	err = os.Chmod(c.GetCachePath(), 0755)
+	if err != nil {
+		return fmt.Errorf("failed to change permissions of cache path %s: %v", c.GetCachePath(), err)
+	}
+
 	c.currentVersion = version
 
+	logger.Info(ctx, c.GetCachePath())
```
airhorns: 🔥 or add message?

test/cached_csi_test.go
```diff
@@ -187,3 +231,26 @@ func TestCachedCSIDriverProbeFailsUntilPrepared(t *testing.T) {
...
+	socket := path.Join(tmpDir, "csi.sock")
+	endpoint := "unix://" + socket
+
+	go func() {
+		err := s.Serve(endpoint)
+		require.NoError(tc.T(), err, "CSI Server exited")
+	}()
+
+	return cached, endpoint, func() { closeClient(); s.Grpc.Stop() }
+}
```
airhorns, line 312: Worth adding a test that proves that the inner DL client can rebuild using the overlay-powered cache presented to it? I am mostly just thinkin there may be other sneaky overlayfs options we need to add

airhorns (approved):
Code LGTM

result
udnay: Oh yeah, I will 🔥 that.

pkg/api/cached.go
udnay: Slipped through, will remove it.

test/cached_csi_test.go
udnay: Sure, I can add that.

udnay: Added above

PR #107, opened 2025-02-24
Add update subpath

udnay:
Adding a `subpath` to object queries. This is similar to kuberentes' subpath construct for [volume mounts](https://kubernetes.io/docs/concepts/storage/volumes/#using-subpath) where a subpath can be used to specify a particular directory within the volume to mount. The same idea applies here where a set subpath prefixes can be used to narrow down the query to rebuild or update just the objects in the selected subpath and not the whole project.


scott-rc (approved):
LGTM 👍

internal/db/queryBuilder.go
```diff
@@ -105,7 +106,12 @@ func (qb *queryBuilder) updatedObjectsCTE() string {
 		ignoresPredicate = "AND o.path NOT LIKE ALL(__ignores__::text[])"
 	}
 
-	return fmt.Sprintf(template, isCachedSelector, cacheJoin, pathPredicate, ignoresPredicate)
+	subpathsPredicate := ""
+	if len(qb.objectQuery.Subpaths) > 0 {
+		subpathsPredicate = "AND o.path LIKE ALL(__subpaths__::text[])"
+	}
```
airhorns, line 112: I guess this option isn't mutually exclusive with `ignores` right -- you could use both at once? And we'll just let PG decide which conditions to use the index for?

airhorns (approved):

internal/db/queryBuilder.go
udnay: You can use both at once, ignoring a subset of the files under the subpaths. I initially wanted to use it like (haskell's lens)[https://github.com/ekmett/lens/wiki/Overview] functionality where the subpaths focus you on a specific set of the project and let you manipulate just that. With that in mind I thought keeping the ignores made sense.

PR #108, opened 2025-02-25
Fixup prerelease

udnay:
Consolidated the pre-release process into a single ts script that keeps all of the different version tags in sync and plays nice with how the gadget code base expects it. This aligns with the versioning of an actual release.

This pulls the `prerelease` changes out of #107 so that it's clearer.


scott-rc (approved):
Nice!

airhorns (approved):

PR #111, opened 2025-03-20
Batch up all deletes/updates in an Update API call

### LINKED ISSUE #106: Improve performance of large updates
We've been seeing some slower updates recently. These are updates for some pretty substantial application file systems (approximately 1.8G), but we still would expect Dateilager (DL) to be able to update a filesystem of that size in under 30 seconds (if we could get that closer to 10 seconds, or even less, that would be amazing!). It looks kind of like this:

Screenshot of a slow trace

So what can we do in DL? From the above, this is the slow codepath that made up the slow span:

https://github.com/gadget-inc/dateilager/blob/140e3dceee212f546ef49720ad1b00ada3daa3ef/pkg/api/fs.go#L627-L647

The first line that jumps out to me is:

```go
for parent, objects := range packedBuffer {
```

Are there a lot of objects? I think so, but we should investigate that. If there are lots, is there any way to batch the DB queries we make, so there's just one big `SELECT` and a bulk `INSERT`, instead of one per object.

Given the above trace, I think that's what's making the `Update` RPC so slow.

If there aren't lots of objects, we'll need to investigate this a bit more to understand why there isn't a lot.

thegedge:
Closes https://github.com/gadget-inc/dateilager/issues/106

Previously, when an update call was issued to dateilager, we would run a set of queries for every object being updated. This resulted in A LOT of DB chatter. For example, during application creating in Gadget:

CleanShot 2025-03-20 at 10 25 50@2x

You can see 1000s of pgx spans are nested under `update-objects` and `update-packed-objects`. This PR fixes that by bulk inserting. Since we have a dynamic number of objects to insert, we can't do a regular INSERT due to potential limitations on the number of parameters. Instead, we use another trick: create a temporary table that is dropped after the transaction completes where we stage all of the updates, and use `COPY` to efficiently get data into it.

The end result is pretty good!

CleanShot 2025-03-20 at 10 29 20@2x

Ignoring the packed objects update (discussion below), the regular update call goes from 1k+ queries taking 289ms to 6 calls taking 44ms. I'll have to benchmark a bunch, but on average I'm thinking this could be close to an order of magnitude improvement. Hard to talk specifics, but it's also likely going to be a big win for the production DB too, since it has to deal with less chatter.

## Topics of conversation

First, the regular update works really well, but I still need to investigate what's making the packed objects version so slow. It's specifically the `CopyFrom` call, but I would have suspected that to be significantly faster.

Second, there were some comments about doing certain things outside of a transaction to avoid deadlocks. I don't do that any more, but I suspect we'll be okay because we do everything in single statements. The comments mentioned deadlocks were observed in tests, but I didn't observe any issues with a Gadget-side PR using a DL built from this branch for testing.

Finally, we likely now need to load a bunch more stuff into memory at once. Assuming I can make the packed objects query have gains similar to the update, how do we feel about that? If we're feeling a bit uncomfy, I can set it all up to batch `CopyFrom` calls so that we stay within some amount of memory per call.

### DIFF
--- a/pkg/api/fs.go
+++ b/pkg/api/fs.go
@@ -509,7 +509,7 @@ func (f *Fs) GetUnary(ctx context.Context, req *pb.GetUnaryRequest) (*pb.GetUnar
 	return &response, nil
 }
 
-// nolint:gocyclo // we should try to break this one up, but ignoring for now
+//nolint:gocyclo // we should try to break this one up, but ignoring for now
 func (f *Fs) Update(stream pb.Fs_UpdateServer) error {
 	ctx := stream.Context()
 
--- a/pkg/api/fs.go
+++ b/pkg/api/fs.go
@@ -509,6 +509,7 @@ func (f *Fs) GetUnary(ctx context.Context, req *pb.GetUnaryRequest) (*pb.GetUnar
 	return &response, nil
 }
 
+// nolint:gocyclo // we should try to break this one up, but ignoring for now
 func (f *Fs) Update(stream pb.Fs_UpdateServer) error {
 	ctx := stream.Context()
 
@@ -522,9 +523,6 @@ func (f *Fs) Update(stream pb.Fs_UpdateServer) error {
 		return status.Errorf(codes.Unavailable, "FS db connection unavailable: %v", err)
 	}
 	defer close(ctx)
-
-	contentEncoder := db.NewContentEncoder()
-	defer contentEncoder.Close()
 
 	var packManager *db.PackManager
 
@@ -583,6 +581,9 @@ func (f *Fs) Update(stream pb.Fs_UpdateServer) error {
 		return stream.SendAndClose(&pb.UpdateResponse{Version: -1})
 	}
 
+	contentEncoder := db.NewContentEncoder()
+	defer contentEncoder.Close()
+
 	latestVersion := int64(-1)
 	nextVersion := int64(-1)
 	shouldUpdateVersion := false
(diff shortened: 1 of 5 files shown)


pkg/api/fs.go
```diff
@@ -594,28 +594,43 @@ func (f *Fs) Update(stream pb.Fs_UpdateServer) error {
 		nextVersion = latestVersion + 1
 		logger.Info(ctx, "FS.Update[Init]", key.Project.Field(project), key.Version.Field(nextVersion))
 
+		var deletedPaths []string
+		var objectsToUpdate []*pb.Object
 		for _, object := range objectBuffer {
```
udnay, line 605: Maybe not for this PR but I wonder if we can fold this loop into the one above, we seem to receive all objects and then write them to the DB which made sense so that we didn't block the RPC stream to do DB calls but now that we are splitting and appending to multiple arrays here, maybe we do it as we receive the objects and then do the bulk update/deletes?

thegedge: Yep, no problem at all to do that!

airhorns:
@thegedge did you want to get this out?

thegedge:
I'm back and forth, but just kept it lying around to focus on some other things for now.

I was curious what you all thought about this approach. Obviously this creates less back and forth DB chatter, but it is more complicated. With the packed object update being much slower, I think it needs a bit more digging before we ship to prod.

If we are happy with the approach in this PR though, I think the plan would be:
1. Do another run with everything on the latest `main`, gather some traces before and after, and compare to make sure it's overall better.
    * If packed object updates are still slow, dig deeper to learn why and see if it's something we can overcome.
    * If we can't figure out why it's slow, we can always revert that part of this PR, but keep the other update path, which definitely did get faster.
1. Throw it at a DL staging environment first, try to get something a bit more "real world".
3. Ship to prod (it should be no problem rolling back).

Which I would do after getting through some of the more urgent issues I've been debugging.

scott-rc (approved):
LGTM on 🍏 -- feel free to `//nolint:gocyclo` on the complexity lint.

https://golangci-lint.run/usage/false-positives/#nolint-directive:~:text=You%20may%20add,the%20flag%20itself%3A

PR #112, opened 2025-03-20
Update otel dependencies

scott-rc:
DL's spans keep referencing the wrong parent. I wasn't able to figure out why so I just updated our otel dependencies and it stopped happening 🤷 

**Before**
CleanShot 2025-03-20 at 18 54 36@2x

**After**
CleanShot 2025-03-20 at 18 53 44@2x


udnay (approved):

thegedge (approved):
`🤷 🚢 🚀                            🌔 `

PR #113, opened 2025-03-20
Drop otel log events

scott-rc:
I made all our logs end up as span events a long time ago so that we could correlate logs with spans. This never ended up being useful and has a lot of drawbacks: 
1. otel only keeps 128 span events before dropping them
2. these span events need to be kept in memory before their flushed
3. we (gadget) drop all span events in our otel collector so this ends up consuming memory and bandwidth for nothing

### DIFF
--- a/internal/telemetry/telemetry.go
+++ b/internal/telemetry/telemetry.go
@@ -2,16 +2,11 @@ package telemetry
 
 import (
 	"context"
-	"fmt"
-	"math"
-	"reflect"
 	"time"
 
 	"github.com/gadget-inc/dateilager/internal/logger"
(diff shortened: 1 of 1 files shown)


udnay (approved):
:shipit: on 🍏

airhorns (approved):
Is the traceid already in the logs also?

scott-rc:
> Is the traceid already in the logs also?

Yeah that's added by `logger.UnaryServerInterceptor` and `logger.StreamServerInterceptor`: 
- https://github.com/gadget-inc/dateilager/blob/main/internal/logger/grpc.go#L28
- https://github.com/gadget-inc/dateilager/blob/main/internal/logger/grpc.go#L62

PR #114, opened 2025-03-28
Allow the ability to set a user

udnay:
We currently have an issue where the app dir is now owned by root instead of by the api user, this change allows us to pass in the container user to the csi driver and we set the app dir to that.

### DIFF
--- a/pkg/api/cached.go
+++ b/pkg/api/cached.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"strconv"
 	"syscall"
 	"time"
 
@@ -31,6 +32,7 @@ const (
 	CACHE_PATH_SUFFIX = "dl_cache"
 	UPPER_DIR         = "upper"
 	WORK_DIR          = "work"
+	NO_CHANGE_USER    = -1
 )
 
 type Cached struct {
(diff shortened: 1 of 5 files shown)


pkg/api/cached.go
```diff
@@ -239,6 +255,17 @@ func (c *Cached) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolu
 		return nil, fmt.Errorf("failed to change permissions of app directory %s: %v", path.Join(targetPath, "app"), err)
 	}
 
+	// For testing where we're not running as root, we need to chown the app dir via the command line :(
+	if appUser != NO_CHANGE_USER && os.Getenv("RUN_WITH_SUDO") == "true" {
+		err = execCommand("chown", "-R", fmt.Sprintf("%d:%d", appUser, appGroup), path.Join(targetPath, "app"))
+	} else {
+		err = os.Chown(path.Join(targetPath, "app"), appUser, appGroup)
+	}
```
scott-rc: Are we supposed to `chown` if `appUser == NO_CHANGE_USER`? i.e. should we do this instead?

```suggestion
	if appUser != NO_CHANGE_USER {
		if os.Getenv("RUN_WITH_SUDO") == "true" {
			err = execCommand("chown", "-R", fmt.Sprintf("%d:%d", appUser, appGroup), path.Join(targetPath, "app"))
		} else {
			err = os.Chown(path.Join(targetPath, "app"), appUser, appGroup)
		}
	}
```

scott-rc (approved):
One comment, but LGTM 👍

pkg/api/cached.go
udnay: From the go docs if `appUser == NO_CHANGE_USER` then the go call no-ops, but that's more readable, I'll change it.

udnay:
Ping @airhorns

airhorns (approved):

PR #117, opened 2025-04-01
Add option to download specific cache version

scott-rc:
This adds the ability to download and extract a specific cache version instead of always downloading the latest one.

I made `-1` mean "give me the latest version", which is the same thing we do for `rebuild`.

/cc @thegedge


internal/db/query.go
```diff
@@ -23,8 +23,10 @@ var (
 	ErrNotFound = errors.New("resource not found")
 )
 
-type EncodedContent = []byte
-type DecodedContent = []byte
+type (
+	EncodedContent = []byte
+	DecodedContent = []byte
+)
```
scott-rc, line 29: These kinds of changes are from a more strict go formatter setting I have set somewhere... I can revert if you'd like but I thought it's okay to keep.

airhorns (approved):
LGTM!

udnay (approved):
:shipit:

PR #118, opened 2025-04-03
reflinks if available

airhorns:
This makes dateilager use reflinks for linking to items in the cache if the filesystem has reflink support, which APFS does. We don't currently make use of the cache in local development in Gadget, but I think we should, and this'll make it faster! 

We still need to benchmark in prod if XFS reflinking through the overlay is faster than ext4 hardlinking with copying up -- we know that the raw reflink speed is worse than raw hardlink speed, but I have not actually established that it is worse than hardlink + copy through the overlay.

### DIFF
--- a/internal/files/writer.go
+++ b/internal/files/writer.go
@@ -300,22 +300,16 @@ func ReflinkDir(olddir, newdir string) error {
 
 		// For regular files, use reflink
 		if info.Mode().IsRegular() {
-			err = reflinkFile(oldpath, newpath)
+			err = reflinkFile(oldpath, newpath, info.Mode())
 			if err != nil {
 				return fmt.Errorf("reflink %v to %v: %w", oldpath, newpath, err)
 			}
 		} else {
 			// For symlinks and other types, just copy the file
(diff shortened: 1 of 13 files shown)


internal/files/writer.go
```diff
@@ -304,3 +396,24 @@ func WriteTar(finalDir string, cacheObjectsDir string, reader *db.TarReader, pac
 
 	return count, fileMatch, nil
 }
+
+// HasReflinkSupport checks if the given directory supports reflinks.
+// It attempts to create a reflink in the directory and returns true if successful.
+func HasReflinkSupport(dir string) bool {
```
udnay, line 399: /nit Is it worth it to have an env var that we pass in that can short circuit this? Once per rebuild is not that bad.

udnay (approved):
This all looks good. Did you check to make sure that the "upperdir" on an `overlayfs` is empty of regular files after an unmount when using reflinks? It likely is but I wouldn't mind something that proves that assumption.

scott-rc (approved):
LGTM

PR #120, opened 2025-04-04
Set the volatile option for overlay mounts for improved performance

airhorns:
https://docs.kernel.org/filesystems/overlayfs.html#volatile-mount

We don't need any real disk persistence at all in prod -- dateilager-server is the source of truth, so I think we can turn this sucker on. Gadget CI run with this prerelease'd: https://github.com/gadget-inc/gadget/pull/15515

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,7 @@ prerelease:
 ifndef tag
 	$(error tag variable must be set)
 else
-	cd js && npx ts-node dateilager-prerelease.ts -t "$(tag)"
+	npx tsx js/dateilager-prerelease.ts -t "$(tag)"
 endif
 
 prerelease-reset:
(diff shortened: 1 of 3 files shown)


js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@gadgetinc/dateilager",
-  "version": "0.9.3",
+  "version": "0.9.5-pre.018867c",
```
udnay: Just make sure you run `make prerelease-reset` before merging into master.

udnay (approved):
I think this is fine. From the docs it mentions
> Volatile mounts are not guaranteed to survive a crash. It is strongly recommended that volatile mounts are only used if data written to the overlay can be recreated without significant effort.

What happens if a pod is restarted? I guess we get a new csi driver location back and away we go?

airhorns:
> What happens if a pod is restarted? I guess we get a new csi driver location back and away we go?

I don't think anything happens to the volumes when a pod is restarted -- they're mounted when the pod is created, and removed when the pod is terminated, but in the middle, the pod is just chilling and its volumes are just sitting there. If I understand correctly, volatile only matters if the system crashes, and I think the k8s node doesn't come back with the same pods if that happens.

js/package.json
airhorns: thx

scott-rc (approved):

PR #121, opened 2025-04-07
Count number of cached files during rebuild

scott-rc:
We updated and increased the size of our dl cache but we're not sure how much of it is being used. This makes the dl client keep track of how many times it uses the dl cache during rebuilds and adds it to the existing "wrote files" log.

/cc @udnay @thegedge

### DIFF
--- a/internal/key/key.go
+++ b/internal/key/key.go
@@ -9,13 +9,18 @@ import (
 )
 
 const (
-	CacheVersions     = Int64SliceKey("dl.cache_versions")
+	CachePath         = StringKey("dl.cache_path")
 	CacheVersion      = Int64Key("dl.cache_version")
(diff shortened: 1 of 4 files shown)


internal/key/key.go
```diff
@@ -158,7 +159,7 @@ func (isk Int64SliceKey) Attribute(value []int64) attribute.KeyValue {
 type DurationKey string
 
 func (dk DurationKey) Field(value time.Duration) zap.Field {
-	return zap.Duration(string(dk), value)
+	return zap.Float64(string(dk), float64(value.Milliseconds()))
```
airhorns, line 162: 🙏

airhorns (approved):

PR #122, opened 2025-04-08
[GGT-7500] Add optional param to GetUnary to not return content above size limit

Robiathin:


internal/db/query.go
```diff
@@ -144,13 +144,19 @@ func executeQuery(ctx context.Context, tx pgx.Tx, queryBuilder *queryBuilder) ([
 }
 
 func loadChunk(ctx context.Context, tx pgx.Tx, lookup *ContentLookup, dbObjects []DbObject, startIdx int, chunkSize int) ([]DecodedContent, error) {
-	hashes := make(map[Hash]bool, chunkSize)
+	return loadChunkSizeLimited(ctx, tx, lookup, dbObjects, startIdx, chunkSize, -1)
```
airhorns: Could we just use an optional argument to the one function instead of having two functions? Seems like it'd be simpler and that the distinction between the limited version and unlimited version is not important if the unlimited version just calls the limited version, you know?

internal/db/content.go
```diff
@@ -186,10 +191,13 @@ func (cl *ContentLookup) Lookup(ctx context.Context, tx pgx.Tx, hashesToLookup m
 	}
 	defer decoder.Release()
 
-	for hash, isEncoded := range hashesToLookup {
+	for hash, params := range hashesToLookup {
+		if params.IsOversize {
+			continue
```
airhorns, line 196: 👍

internal/db/query.go
```diff
@@ -176,6 +187,10 @@ func loadChunk(ctx context.Context, tx pgx.Tx, lookup *ContentLookup, dbObjects
 type ObjectStream func() (*pb.Object, error)
 
 func GetObjects(ctx context.Context, tx pgx.Tx, lookup *ContentLookup, packManager *PackManager, project int64, vrange VersionRange, objectQuery *pb.ObjectQuery) (ObjectStream, error) {
+	return GetObjectsSizeLimited(ctx, tx, lookup, packManager, project, vrange, objectQuery, -1)
```
airhorns: same double function comment, I think it'd be fine to just have one that takes an optional / defaulted arg

airhorns (commented):
Makes sense to me, just a couple comments!

airhorns (approved):

udnay (approved):

PR #123, opened 2025-04-08
Cached Fixes

scott-rc:
This PR contains the changes I had to make to get cached working in sandbox canary.

1. [Make the version file readable to everyone](https://github.com/gadget-inc/dateilager/pull/123/commits/375ce17dc690d6cd3d59c584181099f1f2fea047)
    
    When dateilager goes to rebuild using the `dl_cache`, it first checks which cache versions are available by reading the contents of `dl_cache/versions`. This file was only readable by the csi driver which caused dateilager to [rebuild with an empty list of cache versions](https://github.com/gadget-inc/dateilager/blob/main/pkg/client/dir.go#L49-L53), meaning it wasn't using the `dl_cache` at all.

2. [Chown dl_cache instead of app directory](https://github.com/gadget-inc/dateilager/pull/123/commits/2a39e11fa5c44d51624ffda6a0abf3cc072cb8bd)
	
	After I made the `dl_cache/versions` change, dateilager started to use the `dl_cache`, however it failed to hardlink anything into the pod due to "operation not permitted" errors. The fix was to make every file in the `dl_cache` owned by the app user. Making all the `dl_cache` files owned by the app user makes #114 unnecessary so this change replaces those changes. This also means the csi driver doesn't need to create or manage the app directory at all.

3. [Recreate symlinks and don't assume pb.TarCached is always a directory](https://github.com/gadget-inc/dateilager/pull/123/commits/4bae7ae35d1f0d76823acd53997f0f10072b5be8)

	Another cause of "operation not permitted" errors was from calling `os.Link` on a symlink. You cannot hardlink a symlink, so this makes us recreate them instead, similar to how we recreate directories instead of hardlinking them.

	Lastly, [I made the size of the local cache much bigger](https://github.com/gadget-inc/dateilager/pull/123/commits/70bf01255565b8306146932357f8d0f8ddee5ab3) to test more kinds of files in the cache and ended up with a cache entry without a subdirectory. [We assumed every `pb.TarCached` was a subdirectory](https://github.com/gadget-inc/dateilager/blob/cf4ba135c675bfd0082c37b886faab1c8ccd2770/internal/files/writer.go#L86-L97) which caused rebuilds to fail so this commit removes that assumption.
	
	<details>
	<summary>Cache Entry w/ Single File</summary>
	
	CleanShot 2025-04-14 at 20 16 39@2x
	
	</details>

---

Here's a follow-up PR in Gadget that's been deployed to sandbox canary:
- https://github.com/gadget-inc/gadget/pull/15596


airhorns (commented):
This all makes sense for me but i think we need tests that test the "there's a symlink in the cache" thing or "there's a tarball thats just a file" thing -- both of those burned us and I think will burn us again in the future unless we regression test against them!

udnay (approved):

scott-rc:
> This all makes sense for me but i think we need tests that test the "there's a symlink in the cache" thing or "there's a tarball thats just a file" thing -- both of those burned us and I think will burn us again in the future unless we regression test against them!

@airhorns I amended [Recreate symlinks and don't assume pb.TarCached is always a directory](https://github.com/gadget-inc/dateilager/pull/123/commits/4bae7ae35d1f0d76823acd53997f0f10072b5be8) with the regression tests.

I reused the `copyFile` function you introduced in #118 for non-regular files, however it turns out `copyFile` follows symlinks so it was copying the destination file instead of the symlink file! I wouldn't have caught that without these new regression tests 🫣

airhorns (approved):
LGTM, would be good to have one little assertion on the version file permissions

PR #124, opened 2025-04-09
Simplify prerelease process

scott-rc:
This makes our prerelease script simpler by always using `v0.0.0-pre.<git-sha>` for our prerelease tags, not updating the `package.json` and `default.nix`, and not adding another "update version" commit.

We only need to update the `package.json` and `default.nix` when we release real versions.

/cc @Robiathin

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -198,6 +198,35 @@ Then build and push the image:
 make upload-container-image version=0.0.x
 ```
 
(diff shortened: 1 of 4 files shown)


udnay (approved):
This is much better!

airhorns (approved):
:shipit:

PR #127, opened 2025-04-16
Add --driver-name-suffix

scott-rc:
This adds a `--driver-name-suffix` flag we can pass to the csi-driver to enable blue/green deploys. This also updates our driver name to `dev.gadget.dateilager.cache` instead of `com.gadget.dateilager.cache` since we don't own `gadget.com`.

### DIFF
--- a/pkg/api/cached.go
+++ b/pkg/api/cached.go
@@ -28,7 +28,7 @@ import (
 )
 
 const (
-	DriverName        = "com.gadget.dateilager.cached"
+	DRIVER_NAME       = "dev.gadget.dateilager.cached"
 	CACHE_PATH_SUFFIX = "dl_cache"
 	UPPER_DIR         = "upper"
 	WORK_DIR          = "work"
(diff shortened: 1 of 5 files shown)


airhorns (approved):
This makes sense to me, however, we can't actually run the two side by side with any guarantee they won't mess with one another because they will put the cache in the same spot right? Should we use this same name as the folder name for the cache too perhaps?

scott-rc:
> This makes sense to me, however, we can't actually run the two side by side with any guarantee they won't mess with one another because they will put the cache in the same spot right? Should we use this same name as the folder name for the cache too perhaps? 

I think we can get around that by passing a different `--staging-path`. I'm gonna test this with dateilager-staging and canary.

udnay:
Is the idea we would then mount volumes with `dev.gadget.dateilager.cached-blue/green` ?

udnay (approved):

scott-rc:
> Is the idea we would then mount volumes with `dev.gadget.dateilager.cached-blue/green` ?

Yes exactly -- that + giving them different `--staging-path`s so they don't clobber each others golden caches.

scott-rc:
I deployed a blue and green version of the csi driver to dateilager-staging. Using separate `--driver-name-suffix` + `--staging-path` worked 👍 

CleanShot 2025-04-16 at 15 00 31@2x

CleanShot 2025-04-16 at 14 44 54@2x

PR #132, opened 2025-05-05
Benchmark LVM

scott-rc:

PR #136, opened 2025-05-22
Remove unused cached grpc api

scott-rc:
This PR deletes `cache.proto` and everything associated with it.
- `cache.proto` was generating a grpc api client and server that only had single method named `PopulateDiskCache`
- `PopulateDiskCache` was only allowed to be called in development/test

The grpc api and client wasn't used anywhere, added a lot of noise to the code, and confused me for the longest time, so we should get rid of it imo.


scott-rc:
This change is part of the following stack:

- #132
    - #136 ◀
        - #134

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

airhorns (approved):

PR #138, opened 2025-05-30
Allow passing explicit subpaths to the grpc client

airhorns:
This adds the option to pass subpaths to the JS client, and cleans up the interface a little to use a nicer options object.

### DIFF
--- a/js/spec/grpc-client.spec.ts
+++ b/js/spec/grpc-client.spec.ts
@@ -23,7 +23,7 @@ describe("grpc client operations", () => {
     expect(result?.content?.toString()).toBe("a v1");
   });
 
-  it("can create and read multiple object", async () => {
+  it("can create and read multiple objects", async () => {
     const projectId = 1337n;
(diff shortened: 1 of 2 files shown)


js/spec/grpc-client.spec.ts
```diff
@@ -41,13 +41,93 @@ describe("grpc client operations", () => {
...
+        path: "foo/a.txt",
+        mode: 0o755n,
+        content: "a v1",
+        size: BigInt(1),
+        deleted: false,
+      },
+    ]);
+  });
+
+  it("can ignore objects in a subpath", async () => {
```
udnay, line 84: Can you add a test if both ignore and subpath are set? I'm not sure what the usecase would be except for nested subpaths. `subpaths: ['foo'], ignores: ['foo/beefy-files']`

udnay: I think I already tested this actually so no biggie.

udnay (approved):

PR #144, opened 2025-07-21
Use LVM instead of Overlay

scott-rc:
This changes cached from using overlay to using [LVM](https://www.man7.org/linux/man-pages/man8/lvm.8.html) and [thin snapshots](https://www.man7.org/linux/man-pages/man7/lvmthin.7.html).

/cc @udnay @Robiathin 

## Why

The overhead of using overlay to present the `dl_cache` to pods was too high, such that `dateilager-client rebuild`'s with the `dl_cache` using overlay were slower compared to `dateilager-client rebuild`'s without the `dl_cache` and no overlay.

LVM and thin snapshots allow us to have the same semantics as overlay with much less overhead.

## How

There's a lot to cover so let's go over some LVM terms first:

- Physical Volume (PV)
  - block device (e.g. `/dev/nvme0n1`, `/dev/sda1`)
- Volume Group (VG)
  - named collection of physical volumes (e.g. `vg_dl_cache`)
- Logical Volume (LV)
  - virtual block device, carved out of a volume group (e.g. `vg_dl_cache/pod-1`)
  - appears to the OS like a normal block device (e.g. `/dev/vg_dl_cache/pod-1`)
- Thin Pool
  - special kind of LV that contains the allocated storage for Thin LVs
  - is not used as a block device; does not appear in `/dev`
- Thin LV
  - virtual block device, like a normal LV, except its storage is allocated on demand from a Thin Pool
  - appears to the OS like a normal block device
- Thin Snapshot
  - a Thin LV that is created in reference to an existing Thin LV, Thin Snapshot, or External Origin. The Thin Snapshot initially refers to the same blocks as the existing LV it referenced.
- External Origin
  - a read-only LV that is used as a snapshot origin for Thin LVs. Unwritten portions of the Thin LVs are read from the external origin.

We use all of these concepts to prepare and present the `dl_cache` to pods.

Cached expects to use at least 2 PVs. One PV is for the `dl_cache` which we call the "Base PV", and one or more PVs for the Thin Pool. For the sake of this explanation, let's assume we have the following block devices that we'll use as our PVs:

- `/dev/base`
- `/dev/thin1`
- `/dev/thin2`

First, let's go over `cached.PrepareBasePV` which is the first step in preparing the `dl_cache`. This step is supposed to be done manually on a temporary virtual machine.

> [!NOTE]
> I'm going to use bash to demonstrate the commands that are used, but in practice this is all done in Go code that execs out to the shell.

```bash
# create the Base PV (i.e. let LVM know that this block device is going to be used as a PV)
pvcreate /dev/base

# create a VG named vg_dl_cache and add the Base PV to it
vgcreate vg_dl_cache /dev/base

# create a linear (normal) LV named base in the vg_dl_cache VG
# - make the LV size equal to 100% of all the PVs in the VG (i.e. make the LV the same size as the Base PV)
lvcreate --type linear --name base --extents 100%PVS -y vg_dl_cache

# we now have a virtual block device at /dev/vg_dl_cache/base
# format the base LV with ext4
mkfs.ext4 /dev/vg_dl_cache/base

# mount the base LV
mkdir -p /mnt/vg_dl_cache/base
mount /dev/vg_dl_cache/base /mnt/vg_dl_cache/base

# download the `dl_cache` into the base LV
# ... this is done in Go code via cached.Client.GetCache ...

# unmount the base LV
umount /mnt/vg_dl_cache/base

# make the base LV read-only and deactivate it so it can later be used as an External Origin
lvchange --permission r /dev/vg_dl_cache/base
lvchange --activate n /dev/vg_dl_cache/base
```

Our Base PV (`/dev/base`) is now prepared. It contains a deactivated read-only LV named `vg_dl_cache/base` that contains the `dl_cache`. This Base PV can now be exported, for example as a GCP image, and then attached to Kubernetes nodes as a secondary disk.

Next, let's go over `cached.Prepare` which runs when cached starts up on each Kubernetes node. Assume the Base PV was already prepared and attached to the node as `/dev/base`.

```bash
# import the Base PV into a new VG named vg_dateilager_cached
vgimportclone -n vg_dateilager_cached /dev/base

# extend the new VG with the Thin Pool PVs
vgextend vg_dateilager_cached /dev/thin1 /dev/thin2

# create a Thin Pool named thinpool in the vg_dateilager_cached VG
# - ensure it only uses the Thin Pool PVs
# - make it take up all the space on the Thin Pool PVs
lvcreate --type thin-pool --name thinpool --extents 100%PVS vg_dateilager_cached /dev/thin1 /dev/thin2
```

Now we have a VG named `vg_dateilager_cached` that contains:

- a deactivated read-only LV named `vg_dateilager_cached/base` that contains the `dl_cache`
- a Thin Pool named `vg_dateilager_cached/thinpool` to be used for future Thin Snapshots

If we pass the `--thinpool-cache-lv-size-kib` flag to cached, it will add a write cache to the Thin Pool to make writes faster. A "write cache" is another special kind of LV that is used to batch writes before flushing them to a backing LV. You can read more about LVM caching [here](https://www.man7.org/linux/man-pages/man7/lvmcache.7.html).

Assuming we passed the `--thinpool-cache-lv-size-kib=2097152` flag to cached, here's what `cached.Prepare` will do after creating the Thin Pool:

```bash
# create a ram disk with 2097152 KiB of size (2GiB)
modprobe brd rd_nr=1 rd_size=2097152

# create a PV for the ram disk
pvcreate /dev/ram0

# extend the vg_dateilager_cached VG with the ram disk PV
vgextend vg_dateilager_cached /dev/ram0

# add the ram disk as a write cache to the Thin Pool
lvconvert --type writecache --cachedevice /dev/ram0 vg_dateilager_cached/thinpool
```

Now cached is ready to present the `dl_cache` to pods. When a pod requests a volume, cached will create a Thin Snapshot from the Base LV and present it to the pod.

Assume a pod requests a volume like the following:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: busybox
spec:
  containers:
    - name: busybox
      image: busybox
      command: ["/bin/sh", "-c", "sleep infinity"]
      volumeMounts:
        - name: gadget
          mountPath: /gadget
  volumes:
    - name: gadget
      csi:
        driver: dev.gadget.dateilager.cached
        readOnly: false
```

Cached will recieve a `NodePublishVolume` request for the volume and do the following:

```bash
# create a thin snapshot named vg_dateilager_cached/pod-1
# - use vg_dateilager_cached/thinpool as the Thin Pool
# - use vg_dateilager_cached/base as the external origin
# - don't skip activation so the snapshot is immediately usable
lvcreate --snapshot --thinpool vg_dateilager_cached/thinpool vg_dateilager_cached/base --name pod-1 --setactivationskip n

# create the mount point
mkdir -p /var/lib/kubelet/pods/.../volumes/kubernetes.io~csi/gadget/mount

# mount the thin snapshot to the mount point
mount /dev/vg_dateilager_cached/pod-1 /var/lib/kubelet/pods/.../volumes/kubernetes.io~csi/gadget/mount
```

When the pod starts up, the `dl_cache` will be mounted at `/gadget/dl_cache`. All writes to `/gadget` will be allocated on demand from the `vg_dateilager_cached/thinpool` Thin Pool. All reads from unwritten portions of `/gadget` will be read from the `vg_dateilager_cached/base` Base LV.

This is the same behavior as when using overlay, except with much less overhead :tada:

When the pod is deleted, the `cached.NodeUnpublishVolume` method will be called and do the following:

```bash
# unmount the thin snapshot
umount /var/lib/kubelet/pods/.../volumes/kubernetes.io~csi/gadget/mount

# remove the thin snapshot
lvremove vg_dateilager_cached/pod-1
```

This will free up any allocated space the pod consumed from the Thin Pool.

## Other notes
- This version of cached has already been shipped and deployed for a few weeks now
- LVM is only available on linux, so I had to spin up a GCP VM and use remote ssh to develop this. I've added docs to `Contributing.md` that explains how to setup a GCP VM so we can continue developing this in the future.
- The size of the Base LV is inherited by every Thin Snapshot, so it's important to resize the Base LV after it's prepared. I've added docs about this to `Contributing.md` as well.


pkg/cached/cached.go
```diff
@@ -2,368 +2,598 @@ package cached
...
-			return nil, fmt.Errorf("failed to unmount overlay at %s: %w", targetPath, err)
+	cacheRootDir := filepath.Join(c.BaseLVMountPoint, "dl_cache")
+	cacheVersions := client.ReadCacheVersionFile(cacheRootDir)
+
+	if cacheVersion == -1 || !slices.Contains(cacheVersions, cacheVersion) {
+		logger.Info(ctx, "downloading cache", key.CacheVersion.Field(cacheVersion))
+		startTime := time.Now()
+
+		var cachedCount uint32
+		cacheVersion, cachedCount, err = c.Client.GetCache(ctx, cacheRootDir, cacheVersion)
```
airhorns, line 272: Do we still even want this functionality? I guess its convenient in tests? Maybe we want a flag that disallows this or something though so it doesnt happen in prod by accident?

pkg/cached/cached.go
```diff
@@ -2,368 +2,598 @@ package cached
...
-				Available: signedFreeBytes,
-				Total:     signedFreeBytes + usedBytes,
-				Used:      usedBytes,
-				Unit:      csi.VolumeUsage_BYTES,
-			},
-		},
-	}, nil
-}
+	logger.Info(ctx, "extending vg with thinpool pvs", key.DeviceGlobs.Field(c.ThinpoolPVGlobs), key.PVs.Field(c.ThinpoolPVs))
+	if err := exec.Run(ctx, "vgextend", append([]string{"--config=devices/allow_mixed_block_sizes=1", c.VG}, c.ThinpoolPVs...)...); err != nil {
```
airhorns, line 348: Do we need the mixed block sizes? I think I saw something about it affecting performance negatively.

pkg/cached/cached.go
```diff
@@ -2,368 +2,598 @@ package cached
...
+		// Use one stripe per thinpool device to maximize performance
+		"--stripes", strconv.Itoa(len(thinpoolPVs)),
 
-	cmd := exec.Command(cmdName, args...)
-	return cmd.Run()
+		// Use a small stripe size for better IO performance on small files
+		"--stripesize", "64k",
+
+		// Pass TRIM/discard commands through to underlying storage
+		"--discards", "passdown",
```
airhorns, line 422: Hm, I don't know enough to say for sure but this seems like it'd actually worse for performance

airhorns (approved):
LGTM after comments (I am sure we're gonna change things around a lot as we optimize further)

pkg/cached/cached.go
scott-rc: Yeah this is very convenient in development and tests. When the Base PV was already prepared before hand, we early return at the top of this function [here](https://github.com/gadget-inc/dateilager/blob/9490837196c64dab95300a5176e799561a44368a/pkg/cached/cached.go#L210-L218), so we don't even make it this far.

Everything in `Prepare` must be idempotent incase of restarts and stuff so we should rely on that instead of a flag imo.

pkg/cached/cached.go
scott-rc: Yeah this is the only way to make a VG contain multiple PVs of different block sizes.

Local NVME SSDs are configured with block sizes of 4k whereas hyper/persistent disks are configured with block sizes of 512. I don't remember the exact steps, but I remember trying to change the block sizes was gonna be a big lift so I punted on that.

pkg/cached/cached.go
scott-rc: I think in theory yes, but from my benchmarking it didn't make a difference so I kept it to help mitigate the "out of disk space" risk.

scott-rc:
I'm going to merge this as-is now.

As Harry mentioned, I'm certain we're going to continue to optimize this in the future, but I'm hoping this is good enough foundation to fix forward.

PR #145, opened 2025-08-20
Add cache to base lv

scott-rc:
This adds a `--read-through-base-pv` flag to make cached add a read-through cache in-front of the base LV. This should make reads from the base LV faster when the read-through PV has a faster disk than the base PV.

This is what I'm hoping to use this flag for:

**Before**
CleanShot 2025-08-20 at 15 02 58@2x

**After**
CleanShot 2025-08-20 at 15 03 31@2x

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Setup LVM devices
         run: |
           sudo truncate -s 5G /var/tmp/lvm-base.img
-          sudo truncate -s 5G /var/tmp/lvm-base-cache.img
+          sudo truncate -s 5G /var/tmp/lvm-read-through-base.img
           sudo truncate -s 5G /var/tmp/lvm-thin1.img
           sudo truncate -s 5G /var/tmp/lvm-thin2.img
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,7 +42,7 @@ jobs:
           echo "DL_BASE_LV_FORMAT=${{ matrix.filesystem }}" >> $GITHUB_ENV
           echo "DL_READ_THROUGH_BASE_PV=$read_through_base_device" >> $GITHUB_ENV
           echo "DL_THINPOOL_PV_GLOBS=$thin1_device,$thin2_device" >> $GITHUB_ENV
-          echo "DL_THINPOOL_CACHE_LV_SIZE_KIB=2097152" >> $GITHUB_ENV
+          echo "DL_WRITE_BACK_THINPOOL_PV_SIZE_KIB=2097152" >> $GITHUB_ENV
 
       - name: Checkout repository
         uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


pkg/cached/cached.go
```diff
@@ -45,7 +46,6 @@ type Cached struct {
 	Client                 *client.Client
 	DriverName             string
 	NameSuffix             string
-	ThinpoolCacheLV        string
```
scott-rc, line 48: You can ignore all of these `ThinpoolCacheLV` deletions. This variable wasn't being used/needed that's why it's gone.

.github/workflows/test.yml
```diff
@@ -29,15 +29,18 @@ jobs:
...
           sudo truncate -s 5G /var/tmp/lvm-thin2.img
 
           base_device=$(sudo losetup --find --show /var/tmp/lvm-base.img)
+          base_cache_device=$(sudo losetup --find --show /var/tmp/lvm-base-cache.img)
           thin1_device=$(sudo losetup --find --show /var/tmp/lvm-thin1.img)
           thin2_device=$(sudo losetup --find --show /var/tmp/lvm-thin2.img)
 
           echo "DL_BASE_PV=$base_device" >> $GITHUB_ENV
           echo "DL_BASE_LV_FORMAT=${{ matrix.filesystem }}" >> $GITHUB_ENV
+          echo "DL_BASE_CACHE_PV=$base_cache_device" >> $GITHUB_ENV
```
airhorns: Is it worth running some tests without this and some with?

pkg/cached/cached.go
```diff
@@ -344,6 +337,21 @@ func (c *Cached) importBasePV(ctx context.Context) error {
 		return fmt.Errorf("failed to import base pv %s: %w", c.BasePV, err)
 	}
 
+	if c.BaseCachePV != "" {
```
airhorns: I get why it is named this but since the whole thing is also a cache, it took me a while to figure out what it meant. Could we call it something clearer like `BaseLocalCache` or `ReadthroughBlockCache` or something like that?

airhorns (approved):
LGTM after comments! We're gonna need some way to verify this is actually being used as we expect in prod but I think the lvm prom exporter likely has something arcane 👍

udnay (approved):
LGTM, one question. How does the cache get cleaned up?

scott-rc:
> LGTM, one question. How does the cache get cleaned up?

The cache is entirely managed by LVM, so I'm not exactly sure. Here's a blurb from the [man page](https://man7.org/linux/man-pages/man7/lvmcache.7.html) about it.

> A read and write hot-spot cache, using the dm-cache kernel
         module.  This cache tracks access patterns and adjusts its
         content deliberately so that commonly used parts of the main LV
         are likely to be found on the fast storage. LVM refers to this
         using the LV type cache.

In our case, the "main LV" is the base LV (pd-ssd) and "fast storage" is the read through base PV (local-ssd).

PR #151, opened 2026-03-04
Chunk large GetCompress responses to avoid gRPC message size limit

scott-rc:
Large packed objects (e.g. node_modules tarballs) can exceed the per-message gRPC size limit, permanently breaking sandboxes that have them. PR #149 bumped the limit as a short-term patch; this PR implements the real fix.

A `continued` field is added to the `GetCompressResponse` proto. When `sendPacked` on the server side produces a packed-object blob larger than 4 MiB, it splits it into consecutive stream messages, setting `continued = true` on all but the final chunk. The client's new `recvFullResponse` helper detects a continued response and reads from the stream until it collects a terminating chunk, then reassembles the full blob before handing it to the caller. The reassembly is applied at both recv sites in `Rebuild` -- the initial recv before the worker goroutine starts, and the per-message recv inside the goroutine loop.

Chunking only applies to packed objects; non-packed responses are sent as a single message unchanged, preserving existing behavior.

Tests cover: server-side chunking of a large packed object (verifying `Continued` flags and that the concatenated bytes decode to the correct TAR), the no-chunk path for small packed objects, and the end-to-end rebuild path where the very first `Recv` returns a continued response.

Related to #149.


pkg/api/fs.go
```diff
@@ -404,12 +409,7 @@ func (f *Fs) GetCompress(req *pb.GetCompressRequest, stream pb.Fs_GetCompressSer
 				return status.Errorf(codes.Internal, "FS get next tar: %v", err)
 			}
 
-			err = stream.Send(&pb.GetCompressResponse{
-				Version:  vrange.To,
-				Format:   pb.GetCompressResponse_S2_TAR,
-				Bytes:    tar,
-				PackPath: packPath,
-			})
+			err = sendPacked(stream, vrange.To, tar, packPath, PackedObjectChunkSize)
```
airhorns, line 412: did you look at how hard it might be to actually stream the tar out of wherever we're reading it from? i think because this is gonna enable much larger tars we may need to be more careful with server side memory usage. i think this PR as is is an improvement so we can ship it but we need to watch carefully and perhaps set up some server-side OOM alerts

test/client_rebuild_test.go
```diff
@@ -573,3 +574,107 @@ func TestRebuildWithSubpaths(t *testing.T) {
...
+	tmpDir := emptyTmpDir(t)
+	defer os.RemoveAll(tmpDir)
+
+	rebuild(tc, c, 1, nil, tmpDir, nil, expectedResponse{
+		version: 1,
+		count:   10,
+	}, nil)
+
+	verifyDir(t, tmpDir, 1, expectedFiles)
+}
```
airhorns, line 680: worth adding a test where the tar is right at the boundary size?

airhorns (approved):
This LGTM after comments.

For rollout, I think we need to make sure we distribute the new client version to all sandboxes first, and then ship the new server. my understanding is that protobufs make that ok, where the server doesn't need to produce an optional field, but its probably worth tophatting to make sure we set it all up right

kirinrastogi (approved):

scott-rc:
This change is part of the following stack:

- #151 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

pkg/api/fs.go
scott-rc: I didn't dig into streaming the tar reads -- the main blocker is that `packObjects` materializes the full tar in `db.go` via `tarPackedObjects`, which collects all rows into a `bytes.Buffer` before returning. Streaming would mean piping rows through a `tar.Writer` directly into the gRPC stream, which touches the DB query pattern and the `sendPacked` interface. I'll try and get that working as a follow-up.

test/client_rebuild_test.go
scott-rc: Fixed in eed5b12

PR #152, opened 2026-03-30
Normalize file permission bits during rebuild

alanko0511:
Two representations of the same file permissions coexist in the `dl.objects.mode` column:

- **420** (octal `0644`) — Go-native `os.FileMode`. This is what `os.Lstat().Mode()` returns for a regular file with `rw-r--r--` permissions. Go strips the Unix file type bits for regular files, since "regular" is the default/zero value in Go's type system.

- **33188** (octal `0100644`) — Raw Unix `stat.st_mode`. This includes `S_IFREG` (32768) + the permission bits (420). `S_IFREG` is the Unix constant that tags a file as "regular" in the `stat.st_mode` bitmask.

### What is `S_IFREG`?

On Unix, the `st_mode` field from `stat()` packs two things into one integer:
- **File type** (upper bits): `S_IFREG` (regular file, `0100000`), `S_IFDIR` (directory, `0040000`), `S_IFLNK` (symlink, `0120000`), etc.
- **Permissions** (lower 12 bits): read/write/execute for owner/group/other, plus setuid/setgid/sticky

So `0100644` = `S_IFREG` | `rw-r--r--` = "this is a regular file with permissions 644". The `S_IFREG` part doesn't affect what you can do with the file — it's just a type tag. Go's `os.FileMode` uses its own bit layout and strips `S_IFREG` for regular files (regular is the default), which is why `os.Lstat().Mode()` returns just `0644` (420) instead of `0100644` (33188).

### The problem

Both values have identical permission bits: `33188 & 0o7777 == 420 & 0o7777 == 0o644`.

During incremental rebuild, if a file was previously written to disk with mode 420 and the database now stores 33188 (or vice versa), `writeObject` sees a numeric mismatch (`info.Mode() != os.FileMode(header.Mode)`) and attempts a `chmod`. This unnecessary `chmod` fails with `EPERM` on restricted sandbox filesystems, which kills the entire rebuild and prevents the sandbox from booting.

## Summary

- Mask file type bits (using `& 0o7777`) when comparing on-disk mode vs stored mode during rebuild, so `0644` and `0100644` are recognized as equivalent
- Also mask when passing mode to `os.OpenFile` and `file.Chmod`, so only permission bits are applied
- Add seven tests covering: S_IFREG mode handling, mode representation changes in both directions (perm-only → S_IFREG and vice versa), real permission changes across same and mixed representations, and 0755 mode normalization

### DIFF
--- a/internal/files/writer.go
+++ b/internal/files/writer.go
@@ -101,6 +101,15 @@ func writeObject(rootDir string, cacheObjectsDir string, reader *db.TarReader, h
 		dir := filepath.Dir(path)
 		createdDir := false
 
+		// Mask to extract only permission + special bits (lower 12 bits of Unix
+		// st_mode: rwx for owner/group/other + setuid/setgid/sticky), ignoring
+		// file type bits like S_IFREG (0100000). The dl.objects.mode column
+		// stores both Go-native modes (e.g. 0644 = 420) and raw Unix stat modes
+		// (e.g. 0100644 = 33188) — both represent the same permissions, just
+		// with different representations.
+		const permBits = os.FileMode(0o7777)
+		fileMode := os.FileMode(header.Mode) & permBits
+
 		if _, exists := existingDirs[dir]; !exists {
 			_, err := retryFileErrors(dir, func() (interface{}, error) {
 				createdDir = true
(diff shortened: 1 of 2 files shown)


scott-rc (approved):
I used a similar approach when in ggt to sync file permissions and I remember Alex saying it was a good idea so let's do it here too 👍

